PR #10, opened 2026-01-29
Fix the framework build so it can be referenced in Xcode

lucasnewman:
A few things got busted in the move from the other repo. This fixes the primary framework target to build properly again. Once this is merged I can have the example projects reference it so we can remove the duplicated code.


Blaizzy (approved):
LGTM!

PR #15, opened 2026-02-01
Add a basic CLI tool for TTS generation

lucasnewman:
I also centralized model loading for the TTS models so they have a standardized generation interface. They still call `.fromPretrained()` but one step at a time 😅

Note that I couldn't get the Qwen3-TTS models to work yet, but I want to have the ability to test the models directly (which this does) before I start fixing them up.

### DIFF
--- a/Sources/mlx-audio-swift-tts/App.swift
+++ b/Sources/mlx-audio-swift-tts/App.swift
@@ -37,7 +37,10 @@ enum App {
                 model: args.model,
                 text: args.text,
                 voice: args.voice,
-                outputPath: args.outputPath
+                outputPath: args.outputPath,
+                maxTokens: args.maxTokens,
+                temperature: args.temperature,
+                topP: args.topP
             )
         } catch {
             fputs("Error: \(error)\n", stderr)
@@ -51,6 +54,9 @@ enum App {
         text: String,
         voice: String?,
         outputPath: String?,
+        maxTokens: Int,
+        temperature: Float,
+        topP: Float,
         hfToken: String? = nil
     ) async throws {
         Memory.cacheLimit = 100 * 1024 * 1024
@@ -81,7 +87,11 @@ enum App {
             refAudio: nil,
             refText: nil,
             language: nil,
-            generationParameters: GenerateParameters()
+            generationParameters: GenerateParameters(
+                maxTokens: maxTokens,
+                temperature: temperature,
+                topP: topP
+            )
         ).asArray(Float.self)
 
         let outputURL = makeOutputURL(outputPath: outputPath)
@@ -128,11 +138,13 @@ enum App {
 enum CLIError: Error, CustomStringConvertible {
     case missingValue(String)
     case unknownOption(String)
+    case invalidValue(String, String)
 
     var description: String {
         switch self {
         case .missingValue(let k): "Missing value for \(k)"
         case .unknownOption(let k): "Unknown option \(k)"
+        case .invalidValue(let k, let v): "Invalid value for \(k): \(v)"
         }
     }
 }
@@ -142,12 +154,18 @@ struct CLI {
     let text: String
     let voice: String?
     let outputPath: String?
+    let maxTokens: Int
+    let temperature: Float
+    let topP: Float
 
     static func parse() throws -> CLI {
         var text: String?
         var voice: String? = nil
         var outputPath: String? = nil
         var model = "mlx-community/Qwen3-TTS-12Hz-0.6B-Base-8bit"
+        var maxTokens: Int = 1200
+        var temperature: Float = 0.7
+        var topP: Float = 0.9
 
         var it = CommandLine.arguments.dropFirst().makeIterator()
         while let arg = it.next() {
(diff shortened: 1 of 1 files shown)


Blaizzy:
> I also centralized model loading for the TTS models so they have a standardized generation interface. They still call .fromPretrained() but one step at a time 😅


This is really awesome! I was actually using the tests to models, but this is much cleaner 🔥. I have and can share the schematics of the model factory to use a similar interface to python using Load and post_load_hook. Just didn't get time to implement with all the new models coming out this Jan. 

> Note that I couldn't get the Qwen3-TTS models to work yet, but I want to have the ability to test the models directly (which this does) before I start fixing them up.

No worries, I'm pretty sure we'll have Qwen3-TTS working this coming week.

Sources/mlx-audio-swift-tts/App.swift
Blaizzy, line 1: There a few essential parameters missing here: 

- Lang code 
- Max tokens 
- Temperature
- TopK/P
- Ref Audio 
- Ref text 
- Instructions (for qwen3-tts)

Sources/MLXAudioTTS/TTSModelUtils.swift
```diff
@@ -0,0 +1,77 @@
...
+        modelRepo: String,
+        modelType: String?
+    ) async throws -> SpeechGenerationModel {
+        let resolvedType = normalizedModelType(modelType) ?? inferModelType(from: modelRepo)
+        guard let resolvedType else {
+            throw TTSModelUtilsError.unsupportedModelType(modelType)
+        }
+
+        switch resolvedType {
+        case "qwen3_tts", "qwen_tts", "qwen3", "qwen2_tts", "qwen":
```
Blaizzy: There is no qwen2_tts or qwen_tts

Sources/MLXAudioTTS/TTSModelUtils.swift
```diff
@@ -0,0 +1,77 @@
...
+            return try await Qwen3Model.fromPretrained(modelRepo)
+        case "llama_tts", "llama3_tts", "llama3", "llama", "orpheus", "orpheus_tts":
```
Blaizzy, line 48: Same here

Sources/MLXAudioTTS/Generation.swift
Blaizzy, line 1: Could you elaborate why do we need to extend and duplicate generate in this file?

Sources/mlx-audio-swift-tts/App.swift
lucasnewman: The sampling params are already covered by the GenerateParameters, which is reused from MLXLM. I can add ref audio/text and language to the protocol.

```
/// Parameters for text generation, see ``TokenIterator``.
///
/// This produces:
///
/// - ``LogitSampler``
/// - ``LogitProcessor``
///
/// for the `TokenIterator`.
public struct GenerateParameters: Sendable {

    /// Step size for processing the prompt
    public var prefillStepSize: Int

    /// Maximum tokens to generate
    public var maxTokens: Int?

    /// Maximum size of the key-value cache. Old entries (except the first 4 tokens) will be overwritten.
    /// When set, uses ``RotatingKVCache`` instead of ``KVCacheSimple``
    public var maxKVSize: Int?

    /// Number of bits to use for KV cache quantization. nil implies no cache quantization.
    public var kvBits: Int?

    /// Group size for KV cache quantization (default: 64)
    public var kvGroupSize: Int

    /// Step to begin using a quantized KV cache when kvBits is non-nil (default: 0)
    public var quantizedKVStart: Int

    /// sampling temperature
    public var temperature: Float

    /// top p sampling
    public var topP: Float

    /// penalty factor for repeating tokens
    public var repetitionPenalty: Float?

    /// number of tokens to consider for repetition penalty
    public var repetitionContextSize: Int
```

Sources/MLXAudioTTS/Generation.swift
lucasnewman: It's a protocol, which is how you use a generic interface onto (potentially) multiple concrete implementations in Swift. In python this contract is implicit and resolved at runtime, in Swift it's explicit and resolved at compile time.

lucasnewman: @Blaizzy Take a look now, I tried to make it clearer with models implementing the conformance inline.

Blaizzy: Thanks, I love this new approach more! Easier to maintain and spot issues 💪🏾

Sources/mlx-audio-swift-tts/App.swift
Blaizzy: Makes sense — thanks for the update!

One more thing (and apologies if I'm overlooking something obvious): I noticed that the new arguments are currently passed as nil in the generate call. 

Since this is a CLI tool, would it make sense to wire those up to user input and fall back to nil as the default? Just want to make sure we're not leaving functionality on the table.

lucasnewman: Ok, I added the sampling params. I'll do ref audio/text in a follow up -- I need a working model first to make sure it's actually working correctly.

Blaizzy: Thanks 🚀

PR #17, opened 2026-02-02
feat(tts): add support for Qwen3 TTS

smdesai:
This is a MLX Swift port of mlx_audio/tts/models/qwen3_tts based on an older commit of mlx-audio-swift. 

@Blaizzy As it stands, this code works as-is and has audio quality issues when merged with main. Perhaps the issue is related to Package.swift, I'm trying to determine the cause but if you can shed some light on this, it'll be truly appreciated.

It's using models from smdesai in HF, e.g. smdesai/Qwen3-TTS-12Hz-0.6B-CustomVoice-8bit as the ones in mlx-community are missing tokenizer.json which is required. Once those are present, the model names can be reverted to the ones in mlx-community.

I added a separate target in Voices.app for the model, again, this can be modified to whatever makes sense. 

Appreciate you looking at the implementation for correctness and in particular the issue around audio quality when merging with main.


lucasnewman:
@smdesai Thanks, this looks generally very nice. A couple of notes:

It looks like the stuff in Common isn't really shared -- can you just consolidate it into the Qwen3TTS model code? In practice most architectures don't faithfully share implementations and I think it's both more understandable and useful if the code for any given model is co-located, e.g. when cutting the framework down to run a single specific model in an app.

It also looks like you included some drive-by fixes for deprecation warnings in other models. Those look fine, but can you include them in a separate PR which makes it easier to bisect / roll back if we run into issues?

Also, we should probably incorporate any further changes to Qwen3 TTS from the python repo if you based this on an earlier commit, unless you're planning to do that as a follow up?

smdesai:
> @smdesai Thanks, this looks generally very nice. A couple of notes:
> 
> It looks like the stuff in Common isn't really shared -- can you just consolidate it into the Qwen3TTS model code? In practice most architectures don't faithfully share implementations and I think it's both more understandable and useful if the code for any given model is co-located, e.g. when cutting the framework down to run a single specific model in an app.
> 
> It also looks like you included some drive-by fixes for deprecation warnings in other models. Those look fine, but can you include them in a separate PR which makes it easier to bisect / roll back if we run into issues?
> 
> Also, we should probably incorporate any further changes to Qwen3 TTS from the python repo if you based this on an earlier commit, unless you're planning to do that as a follow up?

@lucasnewman I'll make changes based on the recommendations above.

lucasnewman:
> @lucasnewman, @Blaizzy
> 
> * I've moved the code from Common to Qwen3TTS
> * Reverted the model changes
> * Rebased with main
> * Using Swift Huggingface to interface with the hub not Swift transformers
> 
> What's left to do is updating the VoicesApp view with support instructions so there isn't separate QwenTTS code. I'll have to think about doing this cleanly.
> 
> Now the bad news. With the rebase to main, the models run fine on macOS (M3 Max) but fail on iOS with some words in a phrase not spoken given the impression of fragmented speech. This worked before on the original PR which in its Package.swift was using blaizzy/mlx-swift-lm.
> 
> dependencies: [ .package(url: "https://github.com/ml-explore/mlx-swift.git", branch: "main"), .package(url: "https://github.com/Blaizzy/mlx-swift-lm.git", branch: "main"), .package(url: "https://github.com/huggingface/swift-transformers", .upToNextMinor(from: "1.1.0")), .package(url: "https://github.com/Blaizzy/swift-huggingface.git", branch: "main") ]
> 
> That fork is 27 commits behind mlx-swift-lm and perhaps changes have gone into the ml-explore/mlx-swift-lm repo related to Metal (?) that may be causing problems on iOS. I thought the problem was related to M-series vs. A-series but it seems that on an iPad Pro (M5), I'm seeing the same errors as iOS. Any help to identify the root cause would be appreciated.

We can take a look at the issue there, but w.r.t. your PR, there are still a bunch of things we don't want to merge:

- The model changes for other models are still in the PR
- There's a bunch of .tldr references that looks like it's from your local environment

Can you make sure it looks correct in the PR change listing before you mark it for review again?

PR #19, opened 2026-02-03
Add Pocket TTS

lucasnewman:
I added a README with an example of how to use the model and the supported voices.

### DIFF
--- a/Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSModel.swift
+++ b/Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSModel.swift
@@ -253,9 +253,9 @@ public final class PocketTTSModel: Module, SpeechGenerationModel, @unchecked Sen
     public func generate(
         text: String,
         voice: String?,
-        refAudio: MLXArray?,
-        refText: String?,
-        language: String?,
+        refAudio: MLXArray? = nil,
+        refText: String? = nil,
+        language: String? = nil,
         generationParameters: GenerateParameters
     ) async throws -> MLXArray {
         _ = refText
--- a/Sources/MLXAudioTTS/Models/PocketTTS/README.md
+++ b/Sources/MLXAudioTTS/Models/PocketTTS/README.md
@@ -30,9 +30,6 @@ let model = try await PocketTTSModel.fromPretrained("mlx-community/pocket-tts")
 let audio = try await model.generate(
     text: "Hello world.",
     voice: "alba",
-    refAudio: nil,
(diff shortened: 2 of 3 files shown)


Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSConditioners.swift
```diff
@@ -0,0 +1,72 @@
+import Foundation
+@preconcurrency import MLX
+import MLXNN
+public struct TokenizedText {
+    public let tokens: MLXArray
+}
+
+public final class SentencePieceTokenizer {
```
Blaizzy, line 8: Can we use transformers here?

Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSUtils.swift
```diff
@@ -0,0 +1,74 @@
+import CryptoKit
+import Foundation
+import Hub
+import HuggingFace
+@preconcurrency import MLX
+
+public enum PocketTTSUtils {
+    public static let predefinedVoices: [String: String] = [
```
Blaizzy, line 8: Can't we add this in MLX version in a json or config json? 

That way we don't hardcode and will easily work with any future updates

Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSConditioners.swift
lucasnewman: Sadly no, the Swift UnigramTokenizer they have in place doesn't support SentencePiece with the byte fallbacks, which we need for parity with the python tokenizer.

Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSUtils.swift
lucasnewman: Yeah, I can send a change for that.

Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSConditioners.swift
Blaizzy: Awesome, resolved ✅

Sources/MLXAudioTTS/Models/PocketTTS/PocketTTSUtils.swift
Blaizzy: That would be great! 

Then we can merge

PR #20, opened 2026-02-03
Easier Voices app build

lucasnewman:
Add an .xcconfig template which can be used to set the dev team ID and bundle ID for the app, since the bundle ID is globally unique across all apps and the dev team ID is required for proper code signing for devices. This allows building the Voices app without modifying the project itself. I also added a README with some basic info on the setup.


Blaizzy (approved):
LGTM, thanks a lot for fixing this!

PR #22, opened 2026-02-04
Use voices from the model repo instead of hardcoding

lucasnewman:
@Blaizzy This is the simplified voice loading for Pocket TTS we were discussing, which allows it to be generic and updated from the model repo side.


Blaizzy (approved):
Perfect, LGTM! 🚀

PR #23, opened 2026-02-06
Add Qwen3-TTS conditional generation (VoiceDesign) support

spokvulcan:
## Summary

Adds Qwen3-TTS conditional generation support, ported from the [Python mlx-audio](https://github.com/Blaizzy/mlx-audio/tree/main/mlx_audio/tts/models/qwen3_tts) implementation. This enables text-to-speech with the VoiceDesign path (text + voice description → speech) using models like `Qwen/Qwen3-TTS`.

Related: #17 — this is an alternative, more focused implementation.

## Architecture

**5 new files** in `Sources/MLXAudioTTS/Models/Qwen3TTS/`:

| File | Lines | Purpose |
|------|-------|---------|
| `Qwen3TTSConfig.swift` | ~340 | Config structs (talker, speech decoder, code predictor) |
| `Qwen3TTSModel.swift` | ~590 | Main model + `tokenizer.json` generation from `vocab.json` + `merges.txt` |
| `Qwen3TTSTalker.swift` | ~360 | Talker transformer with 3D multimodal RoPE |
| `Qwen3TTSSpeechDecoder.swift` | ~830 | RVQ decoder with ConvNeXt upsampling pipeline |
| `Qwen3TTSCodePredictor.swift` | ~240 | Multi-codebook token prediction sub-model |

**1 modified file**: `TTSModelUtils.swift` — smart model routing that inspects `config.json` for a nested `talker_config` key to distinguish Qwen3-TTS conditional generation models from base Qwen3 models.

## Key design decisions

- **Flat structure**: All Qwen3-TTS code lives in a single `Qwen3TTS/` directory (5 files), co-located with the existing `Qwen3/` base model. No new top-level directories or scattered utilities.
- **Config-based routing**: `TTSModelUtils` detects `talker_config` in `config.json` to automatically route to `Qwen3TTSFullModel` vs the existing `Qwen3Model`. No model type string changes needed.
- **Tokenizer generation**: Qwen3-TTS HuggingFace repos ship without `tokenizer.json` (only `vocab.json` + `merges.txt`). The model generates the fast tokenizer JSON on first load so `swift-transformers` can use it.
- **Streaming decode**: Speech decoder supports chunked decoding for memory-efficient audio generation on large sequences.

## Differences from #17

| Aspect | This PR | #17 |
|--------|---------|-----|
| Files | 5 new + 1 modified | ~20 files, ~6400 lines |
| Structure | Flat `Qwen3TTS/` directory | Spread across multiple directories |
| Routing | Config inspection (`talker_config` key) | Separate model type strings |
| Tokenizer | Auto-generates `tokenizer.json` | Requires pre-existing tokenizer |

## Test plan

- [x] `swift build` compiles without errors
- [x] Load a Qwen3-TTS model (e.g. `Qwen/Qwen3-TTS`) — should route to `Qwen3TTSFullModel`
- [x] Load a base Qwen3 model — should still route to `Qwen3Model`
- [x] Generate speech with VoiceDesign path

### DIFF
--- a/Examples/SimpleChat/Config/Base.xcconfig
+++ b/Examples/SimpleChat/Config/Base.xcconfig
@@ -1,0 +1,8 @@
+// Shared settings for all builds.
+APP_BUNDLE_ID = com.example.SimpleChat
+PRODUCT_BUNDLE_IDENTIFIER = $(APP_BUNDLE_ID)
+
+// Keep empty so Xcode prompts when needed.
+DEVELOPMENT_TEAM =
+CODE_SIGN_STYLE = Automatic
+CODE_SIGN_IDENTITY = Apple Development
--- a/Examples/SimpleChat/Config/Debug.xcconfig
+++ b/Examples/SimpleChat/Config/Debug.xcconfig
@@ -1,0 +1,2 @@
+#include "Base.xcconfig"
+#include "Local.xcconfig"
--- a/Examples/SimpleChat/Config/Local.xcconfig.template
+++ b/Examples/SimpleChat/Config/Local.xcconfig.template
@@ -1,0 +1,3 @@
+// Copy to Local.xcconfig and customize.
+APP_BUNDLE_ID = com.example.VoicesApp
+DEVELOPMENT_TEAM = YOURTEAMID
--- a/Examples/SimpleChat/Config/Release.xcconfig
+++ b/Examples/SimpleChat/Config/Release.xcconfig
@@ -1,0 +1,2 @@
+#include "Base.xcconfig"
+#include "Local.xcconfig"
--- a/Examples/SimpleChat/Package.swift
+++ b/Examples/SimpleChat/Package.swift
@@ -1,0 +1,27 @@
+// swift-tools-version: 6.2
+
+import PackageDescription
+
+let package = Package(
+    name: "SimpleChat",
+    platforms: [
+        .iOS(.v26),
+        .macOS(.v26)
+    ],
+    products: [
+        .executable(name: "SimpleChat", targets: ["SimpleChat"])
+    ],
+    dependencies: [
+        .package(path: "../..")
+    ],
+    targets: [
+        .executableTarget(
+            name: "SimpleChat",
+            dependencies: [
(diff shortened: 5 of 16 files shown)


Sources/MLXAudioTTS/TTSModelUtils.swift
Blaizzy, line 1: We should have separate paths  qwen3 and qwen3_tts because they are different.

I know it's a bit complex right now but we are stil refining the loader to catch models by model_type or name.

Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
Blaizzy, line 1: we can call this Qwen3TTS.swift

Blaizzy:
The goal of this should be to be a complete PR. Including all qwen3 features :)

INQTR:
Addressed the initial feedback separated qwen3 and qwen3_tts into distinct switch cases, renamed the file/class, and removed the 6bit model, looking forward to the full review.

One question on sampling parameters I've noticed that with high temperature, the output varies significantly from the voice design description. In a separate version of fork I use for my project, I've tuned the defaults differently than for models with predefined voices. Specifically, I made it so that only changing the voice description parameter (or manually changing the seed) produces a different-sounding voice, this helps achieve consistent generations across chunks.

Guys how would you like to handle the default sampling parameters for this model?

Blaizzy:
> One question on sampling parameters I've noticed that with high temperature, the output varies significantly from the voice design description. In a separate version of fork I use for my project, I've tuned the defaults differently than for models with predefined voices. Specifically, I made it so that only changing the voice description parameter (or manually changing the seed) produces a different-sounding voice, this helps achieve consistent generations across chunks.


Yes, temperature affects the talker/LLM. We could have seed as an argument that users can pass. This way the can create reproducible voices not only across chunk but even generations.

lucasnewman:
@Blaizzy I think we're good to merge here if you're ok with handling the other features in a follow up. I think getting the pieces in here even incrementally is a good step to start building on.

PR #25, opened 2026-02-06
Add a simple local chat example

lucasnewman:
This adds a new minimal example app showing a fully local speech-to-speech loop, where you can talk to a voice assistant and it will respond. It uses the built-in STT and Foundation (LM) models in iOS/macOS 26+ along with Pocket TTS (by default) to keep things on-device.

I had to include a couple of fixes to make this work:

- The swift-huggingface library has a bug where if we don't specify allowed extensions it will try to download directories as files, fail the download and drop a sentinel error file, and then fail out when it tries to download files in the directory 🤦 I put in a workaround to pin to .json and .safetensors files for now.
- I had to port over the longform cache fix for Pocket TTS from python (https://github.com/Blaizzy/mlx-audio/pull/486)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ MLXAudio follows a modular design allowing you to import only what you need:
 
 - **MLXAudioCore**: Base types, protocols, and utilities
 - **MLXAudioCodecs**: Audio codec implementations (SNAC, Vocos, Mimi)
-- **MLXAudioTTS**: Text-to-Speech models (Soprano, Qwen3, LlamaTTS)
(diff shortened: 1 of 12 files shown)


lucasnewman:
> LGTM!
> 
> A couple comments:
> 
> 1. Could you share a reproducible example for Swift-Huggingface issue. The maintainer is pretty fast 👌🏽
> 2. Wonder if in the future we can unify these apps (voices + simpleChat) into one and separate by view?

Yep, I want to have a good model base and API in place and then we can improve the app(s) quite a bit.

PR #30, opened 2026-02-08
Improved TTS sampling param handling

lucasnewman:
This adds sampling defaults per-model for TTS models, with the ability to customize any individual param as needed. This makes it much easier to get correct sampling by default for any given model and reduces the API surface area.

I also updated the command-line app and Voices app to use model defaults, with both still having the ability to override any parameter.


Blaizzy (approved):
LGTM!

PR #45, opened 2026-02-15
Add remaining support for Qwen3 TTS base models

lucasnewman:
This is a minimal/direct port of the remaining modules for the Qwen3 TTS base models from the python codebase. I verified unconditional and voice matching work as expected from the `mlx-audio-swift-tts` cli tool:

```bash
mlx-audio-swift-tts \
--model "mlx-community/Qwen3-TTS-12Hz-0.6B-Base-bf16" \
--text "Hello world from Qwen 3 TTS."  --ref_audio <audio_path> --ref_text <audio_caption>
```


Blaizzy (approved):
LGTM!

btw I loved the self, it helped me quickly see where things come from 😂

PR #46, opened 2026-02-16
Add SAM Audio STS model

lucasnewman:
This adds the SAM Audio implementation as a direct port from the python functionality. I also needed to make some fixes to the DACVAE implementation -- if you want the codec stuff separately just lmk, but I didn't want to create a big waterfall of patches.

I added command line tools both codec-based reconstruction and STS audio segmentation, along with comprehensive tests for the model.


Package.swift
```diff
@@ -100,7 +115,10 @@ let package = Package(
                 .product(name: "HuggingFace", package: "swift-huggingface"),
                 .product(name: "Transformers", package: "swift-transformers"),
             ],
-            path: "Sources/MLXAudioSTT"
+            path: "Sources/MLXAudioSTT",
+            exclude: [
+                "Models/GLMASR/README.md",
```
Blaizzy: Let's add pattern match because we are gonna have lots of readme files

Sources/Tools/mlx-audio-swift-codec/App.swift
Blaizzy, line 1: I love the idea but what's confusing to me is the folder organization.

What if we name the parent CLI, Tools or something then have subfolders called codec, sts and so on?

Blaizzy: I saw the tts one, 

It makes sense, but I wonder if we can have a CLI with a parent folder

Maybe `mlx-audio-swift/task` it will be closer to python design and help us structure things better.

Package.swift
lucasnewman: Unfortunately pattern matching isn't supported here (which surprised me too), but the alternative is just a build warning so maybe that's better than needing to keep this up to date.

Sources/Tools/mlx-audio-swift-codec/App.swift
lucasnewman: A 'Tools' directory makes sense, I moved them.

Package.swift
lucasnewman: I reverted those changes.

Blaizzy: Awesome, maybe we move all readme files into docs/model_type/ in the future for swift.

Sources/Tools/mlx-audio-swift-codec/App.swift
Blaizzy: Perfect!

PR #41, opened 2026-02-13
feat: Add support for Voxtral Realtime 4B

shreyaskarnik:
porting https://github.com/Blaizzy/mlx-audio/pull/487 to Swift. 

I am relatively new to swift but tried to follow guidelines in this project. 
Note: I have used Claude Code to understand and implement parts of this PR.

PR #52, opened 2026-02-18
Add Voxtral Realtime STT model

lucasnewman:
A minimal implementation based on https://github.com/Blaizzy/mlx-audio-swift/pull/41. I marked @shreyaskarnik as co-author here since he did a lot of the work.

### DIFF
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
@@ -57,10 +57,6 @@ public final class VoxtralRealtimeModel: Module, STTGenerationModel {
         audio: MLXArray,
         generationParameters: STTGenerateParameters
     ) -> STTOutput {
-        guard tokenizer != nil else {
-            return STTOutput(text: "", language: generationParameters.language)
-        }
-
         let audio1D = audio.ndim > 1 ? audio.mean(axis: -1) : audio
         var context = encodeAndPrefill(
             audio: audio1D,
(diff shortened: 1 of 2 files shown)


Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
```diff
@@ -0,0 +1,541 @@
...
+        self._decoder.wrappedValue = VoxtralRealtimeDecoder(config.decoder)
+    }
+
+    public func generate(
+        audio: MLXArray,
+        generationParameters: STTGenerateParameters
+    ) -> STTOutput {
+        guard tokenizer != nil else {
+            return STTOutput(text: "", language: generationParameters.language)
+        }
```
Blaizzy: I like how graceful it is but shoudn't this error out at load time?

Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeDecoder.swift
```diff
@@ -0,0 +1,276 @@
...
+
+    func embedToken(tokenId: Int) -> MLXArray {
+        tokEmbeddings.weight[tokenId]
+    }
+
+    func embedTokens(_ tokenIds: MLXArray) -> MLXArray {
+        tokEmbeddings(tokenIds)
+    }
+
+    func forward(
```
Blaizzy: can we rename this to the native call instead of forwad?

Blaizzy:
All looks good, just a couple nits before we merge

Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
lucasnewman: Yep, fair.

Blaizzy (approved):
LGTM!

PR #56, opened 2026-02-19
Refactor sts app

Blaizzy:


Sources/MLXAudioSTS/STSModelUtils.swift
```diff
@@ -0,0 +1,96 @@
...
+            return "Invalid repository ID: \(repo)"
+        case .unsupportedModelType(let modelType):
+            return "Unsupported STS model type: \(String(describing: modelType))"
+        }
+    }
+}
+
+// MARK: - Factory
+
+public enum STSModelUtils {
```
lucasnewman: What do you think about making this an extension on STSModel instead, so callers can do `STSModel.loadModel(<repoID>)` and get back the correctly instantiated model? I think it reads a bit better (it's more like AutoModelForX in transformers) and we could make the same change for the TTS models. It's probably better to make the API changes now than before there's a release if we want to do it.

ekryski: As someone coming out of nowhere to you guys but who has been using this lib for ~1 month now, I think this is a good change. I have a fork I'm needing to run and maintain because some of the model methods aren't public enough and things like caching location are fixed.

Also working on chatterbox port and have Kokoro ported and a few other minor fixes. Will try and separate them into nice clean commits and put up some PRs this week.

Also nice work on getting Voxtral in @lucasnewman! Hoping to take it for a spin today.

lucasnewman: Yes, the caching issue is one I plan to fix! I have a change for it locally but it needs a fair amount of testing, but I think you can rely on the idea that we'll have developer-specifiable cache paths coming soon.

Blaizzy: Makes sense, I completely agree!

That was the objective

Blaizzy: > As someone coming out of nowhere to you guys but who has been using this lib for ~1 month now, I think this is a good change.
> Also working on chatterbox port and have Kokoro ported and a few other minor fixes. Will try and separate them into nice clean commits and put up some PRs this week.

@ekryski really happy to hear that! 
We are gearing towards official launch we are trying to get close parity with our python SDK so model PRs are welcome. Chatterbox and chatterbox-turbo are great additions on the backlog.


>  I have a fork I'm needing to run and maintain because some of the model methods aren't public enough and things like caching location are fixed.

Could you point those out in a issue with use case and reproducible examples? 
It will help us think about how to adjust our API design.

Blaizzy: @lucasnewman refactored to STSModel.load()

ekryski: Sounds good! FWIW I have a WIP branch for Chatterbox here: https://github.com/ekryski/mlx-audio-swift/tree/ek/chatterbox. Needs a bunch of clean up but I'm working on it right now.

I (Claude and I) finally just got chatterbox-turbo working properly and tested with it's default voice. Took a while because I didn't realize that turbo and the larger chatterbox models have fundamentally different architectures.

I'm testing out turbo and regular chatterbox-tts with reference voices now. Once I have confirmed that is working well, I'll do a clean up and optimization pass and put up a PR.

One final comment I will say is that I have a benching suite I created in my app that utilizes these models. Parakeet V2 and my Kokoro port are still the fastest. I also have some optimizations that I think might produce better streaming results, especially for Parakeet models because they don't natively support streaming.

@Blaizzy my goal here is to create separate branches for each of these models and put up PRs for them. And possibly create another PR to move my benchmarking suite over to `mlx-audio-swift` as well. I think more people can benefit from being able to see speed vs. WER vs. memory and CPU usage across the various models. Let me know if that's something you're interested in and I can create some issues to track.

Issue #54, opened 2026-02-18
Support configurable model cache directory


BenRacicot:
Lovin' MLX!!! I found a show stopper I wanted to create an issue about.

  ## Summary

  The library hardcodes the model cache directory to `~/Library/Caches/mlx-audio/<org>_<model>/` with no way to override it. It would
  be very useful to support a configurable base cache directory.

  ## Current Behavior

  The cache path is hardcoded independently in ~8 different model resolve functions across the codebase:

  - `ModelUtils.resolveOrDownloadModel()` in MLXAudioCore — uses `URL.cachesDirectory`
  - `Soprano.swift` — duplicates with `FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)`
  - `LlamaTTS.swift` — same duplication
  - `PocketTTSModel.swift` — same duplication
  - `Qwen3.swift` — same duplication
  - `SNACDecoder.swift` — same duplication
  - `GLMASR.swift` — same duplication
  - `MarvisTTSModel.swift` — separate prompt cache path

  None of these expose any mechanism to configure the directory — no environment variable, no constructor parameter, no protocol.

  ## Problem

  As a host app embedding mlx-audio-swift, I'd like to:

  1. **Unify model storage** — my app embeds multiple inference libraries (mlx-audio-swift, FluidAudio, llama.cpp), each with its own
  hardcoded cache location. I'd like all models in one managed directory for a better user experience (one place to see disk usage,
  one place to clean up).

  2. **Control disk usage** — without a configurable path, there's no clean way to manage where potentially large model files
  (multi-GB) end up.

  3. **Check cache status** — since the path logic is internal, host apps that need to check if a model is already downloaded have to
  reverse-engineer the internal path convention (replace `/` with `_`, check for `.safetensors` files, etc.).

Thanks! And let me know if I can help at all.

-Ben

  ## Suggested Approach

 The HuggingFace Python ecosystem supports `HF_HUB_CACHE` for this purpose. Parity with that convention would benefit any app embedding this library.

  Or maybe a centralized configuration point:

  ```swift
  // Option A: Static configuration
  MLXAudioConfig.cacheDirectory = URL(...)

  // Option B: Environment variable
  // Respect MLX_AUDIO_CACHE_DIR if set, fall back to current default

  // Option C: Parameter on load functions
  let model = try await TTSModelUtils.loadModel(
      modelRepo: "mlx-community/Soprano-1.1-80M-bf16",
[...]

PR #60, opened 2026-02-20
Allow configuring the cache location for downloaded models

lucasnewman:
This updates all models to pass through `ModelUtils.resolveOrDownloadModel()` and adds a cache parameter to all `.fromPretrained()` functions to allow the developer to specify a cache location.

I also updated the TTS interface to use the same simplified factory pattern as STS, with some deprecation notes to allow existing clients to still compile in the meantime with a fix-it in Xcode to update to the latest API.

Resolves https://github.com/Blaizzy/mlx-audio-swift/issues/54

### DIFF
--- a/Examples/VoicesApp/VoicesApp/ViewModels/TTSViewModel.swift
+++ b/Examples/VoicesApp/VoicesApp/ViewModels/TTSViewModel.swift
@@ -31,6 +31,7 @@ class TTSViewModel {
     private var maxTokensOverride: Int?
     private var temperatureOverride: Float?
     private var topPOverride: Float?
+    private var repetitionPenaltyOverride: Float?
 
     var maxTokens: Int {
         get { maxTokensOverride ?? defaultMaxTokens }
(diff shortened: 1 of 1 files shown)


Sources/MLXAudioTTS/Models/Marvis/MarvisTTSModel.swift
```diff
@@ -101,7 +101,10 @@ public final class MarvisTTSModel: Module {
     
     private func cacheURL(for audioURL: URL) -> URL {
         let cacheFilename = audioURL.deletingPathExtension().appendingPathExtension("npy").lastPathComponent
-        let cacheDirectory = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!.appending(component: "MarvisTTSModel/prompt_cache")
+        let cacheDirectory = HubCache.default.cacheDirectory
+            .appendingPathComponent("mlx-audio")
+            .appendingPathComponent("MarvisTTSModel")
+            .appendingPathComponent("prompt_cache")
```
Blaizzy: This is the only model hardcoded.

I like this approach better:

```
 let modelSubdir = repoID.description.replacingOccurrences(of: "/", with: "_")

 let modelDir = cache.cacheDirectory
            .appendingPathComponent("mlx-audio")
            .appendingPathComponent(modelSubdir)
```

Blaizzy: Also, what is the prompt cache and .npy files we are saving?

Blaizzy (dismissed):
LGTM, just have one nit

Sources/MLXAudioTTS/Models/Marvis/MarvisTTSModel.swift
lucasnewman: It's a Mimi token cache, but in practice this isn't used so I'll just delete it.

PR #64, opened 2026-02-21
Add Smart Turn v3 model

lucasnewman:
I also added audio resampling as a first-class utility free function, and `loadAudioArray()` now takes an optional sample rate and will automatically resample audio to the target rate.


Sources/MLXAudioVAD/Models/SmartTurn/LICENSE
Blaizzy, line 1: Instead of having the license hardcoded here, we could mention and put a link to it in the readme

Blaizzy (approved):
LGTM!

Just have one tiny nit about the license

Sources/MLXAudioVAD/Models/SmartTurn/LICENSE
lucasnewman: Sure!

PR #68, opened 2026-02-22
Add stock voices and clean up UI in the Voices app

lucasnewman:
Some updates to the Voices app to make it more usable:

1) Add stock voices that match the descriptions in the voice selector
2) Use Qwen3 TTS 0.6B base as the default model
3) Clean up the UI to make it a bit more legible / standard and fix some visual issues
4) Allow pressing command-return to generate without needing to click the Generate button
5) Auto-play generated segments for instant previews


Blaizzy (approved):
LGTM 🚀

PR #74, opened 2026-02-24
Add a CLI option to emit timestamps for speech <-> text alignment

### LINKED ISSUE #73: TTS + forced alignment in single step
[As requested](https://x.com/Prince_Canuma/status/2026263850572562845), I would love to see sample code (or built in API!) to handle generating audio with timestamps in a single pass, similar to how Kokoro works currently.

lucasnewman:
This enables single-step generation + alignment. See https://github.com/Blaizzy/mlx-audio-swift/issues/73

`% mlx-audio-swift-tts --model mlx-community/pocket-tts --voice cosette --text "The quick brown fox jumps over the lazy dog." --timestamps`

```sh
Loading model (mlx-community/pocket-tts)
Generating
Wrote WAV to output.wav
Timestamps:
  [0.560s - 0.560s] The
  [0.560s - 0.800s] quick
  [0.800s - 1.040s] brown
  [1.040s - 1.360s] fox
  [1.360s - 1.680s] jumps
  [1.680s - 1.840s] over
  [1.840s - 1.920s] the
  [1.920s - 2.320s] lazy
  [2.320s - 2.720s] dog
Finished generation in 2.12s
Memory usage:
Peak:   2028M        (2127404412)
Active: 233M         (244890152)
Cache:  100M         (105003252)
Done. Elapsed: 2.12s
```

Closes #73

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,15 @@
+# These are supported funding model platforms
+
+github: Blaizzy # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+lfx_crowdfunding: # Replace with a single LFX Crowdfunding project-name e.g., cloud-foundry
+polar: # Replace with a single Polar username
+buy_me_a_coffee: # Replace with a single Buy Me a Coffee username
+thanks_dev: # Replace with a single thanks.dev username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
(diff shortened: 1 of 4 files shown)


Sources/Tools/mlx-audio-swift-tts/App.swift
```diff
@@ -159,6 +186,14 @@ enum App {
         let audioFile = try AVAudioFile(forWriting: outputURL, settings: format.settings)
         try audioFile.write(from: buffer)
     }
+
+    private static func prepareAudioForForcedAlignment(samples: [Float], sampleRate: Int) throws -> MLXArray {
```
Blaizzy, line 190: Not sure about this function.

Don't we handle this inside resample audio?

lucasnewman: The model itself doesn't resample input audio -- is that what you mean? This is just calling `resampleAudio()` fwiw.

Blaizzy: I meant the resample function should handle all of this logic internally so that we don't need to create custom functions like this one

lucasnewman: Ah, I see. Yeah, it deals in [Float] right now so it's not redundant per se. I can add an MLXArray-based path for `resampleAudio` in a follow up though, I think that would be useful.

Blaizzy: Yes, let's do it! 

I hit the same issue but with the audio player 

We should add a path to keep MLX arrays and only convert when saving

PR #82, opened 2026-03-05
More performance improvements for Qwen3 TTS

lucasnewman:
This brings over the incremental decode from https://github.com/Blaizzy/mlx-audio/pull/537 and adds streaming interval as a supported parameter to allow configuring the chunk size for streaming delivery.

I also added a `--benchmark` flag to the CLI tool to make it easy to measure TTFB/RTF. On my M3 Max with `mlx-community/Qwen3-TTS-12Hz-1.7B-VoiceDesign-bf16`:

```sh
Loading model (mlx-community/Qwen3-TTS-12Hz-1.7B-VoiceDesign-bf16)
Generating
Finished generation in 5.38s
Benchmark:
  Audio duration: 11.60s
  TTFB: 0.229s
  RTF: 0.464
  Tokens/s: 27.06
Memory usage:
Peak:   4889M        (5126549189)
Active: 4338M        (4549253408)
Cache:  256M         (268598075)
```

### DIFF
--- a/Sources/Tools/mlx-audio-swift-tts/App.swift
+++ b/Sources/Tools/mlx-audio-swift-tts/App.swift
@@ -189,10 +189,10 @@ enum App {
                 print("  TTFB: n/a")
             }
             if benchmarkMetrics.audioDuration > 0 {
-                let rtf = benchmarkMetrics.elapsed / benchmarkMetrics.audioDuration
-                print(String(format: "  RTF: %.3f", rtf))
+                let rtf = benchmarkMetrics.audioDuration / benchmarkMetrics.elapsed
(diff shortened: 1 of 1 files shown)


Blaizzy:
I think the RTF calculation is wrong 

It generated 11 seconds of audio in 5 seconds and RTF is still 0.4 🧐

Shouldn't it be around 2.5x and above ?

Blaizzy (changes requested):
LGTM, thanks Lucas! 

The only concern is the RTF calculation, after that's fixed we can merge 🚀

Also, would love to see some metrics on how it performs compared to the Python results. 

I suspect it's now much faster than Python.

PR #88, opened 2026-03-09
add Chatterbox TTS model with voice cloning support

ekryski:
Port of ResembleAI's Chatterbox TTS to MLX Swift. Includes both the Turbo (GPT-2 backbone) and Regular (LLaMA backbone) model variants with full S3Gen vocoder pipeline, T3 text-to-semantic generation, S3 tokenizer, CAMP+ speaker encoder, and HiFT-based neural vocoder.

Features:
- Text-to-speech with 24kHz output
- Voice cloning via reference audio conditioning
- Streaming generation support
- Default conditioning from bundled speaker embeddings
- Emotion tag support ([laugh], [sigh], etc.) via S3Gen pipeline

Tested models:
- mlx-community/chatterbox-turbo-fp16
- mlx-community/chatterbox-turbo-8bit
- mlx-community/chatterbox-turbo-4bit
- mlx-community/Chatterbox-TTS-fp16

> Note: the model works just fine but I'm finding that the audio doesn't match the reference audio file as much as I would expect. They sound close but not as good as the the [ResembleAI Demo Page](https://resemble-ai.github.io/chatterbox_demopage/). This leads me to believe that there is still a bug in the implementation. Happy to see someone else help improve this! 😄

### DIFF
--- a/Sources/MLXAudioSTT/Models/GraniteSpeech/GraniteSpeech.swift
+++ b/Sources/MLXAudioSTT/Models/GraniteSpeech/GraniteSpeech.swift
@@ -1,0 +1,1147 @@
+//
+//  GraniteSpeech.swift
+//  MLXAudioSTT
+//
+
+import Foundation
+import MLX
+import MLXAudioCore
+import MLXNN
+import MLXLMCommon
+import HuggingFace
+import Tokenizers
+
+// MARK: - Language Codes
+
+private let languageCodes: [String: String] = [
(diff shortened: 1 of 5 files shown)


lucasnewman:
@ekryski Can you apply this diff to fix the tests and warnings? https://gist.github.com/lucasnewman/00ac4dfa3eee4abeb088793cfe5be946

Once that's in we can merge.

PR #94, opened 2026-03-11
Add clarifying comments on how to run tests, and fix test artifact remnants

lucasnewman:
I also made sure all the CI tests avoid model downloads, so they should run quickly (~2 mins on my machine).


Blaizzy (approved):
LGTM, thanks!

Much better 🚀

PR #95, opened 2026-03-12
Add granite speech 4

Blaizzy:


Sources/MLXAudioSTT/Models/GraniteSpeech/GraniteSpeech.swift
```diff
@@ -0,0 +1,1147 @@
...
+                if sanitizedWeights["\(path).scales"] != nil {
+                    return perLayerQuantization.quantization(layer: path)?.asTuple
+                }
+                return nil
+            }
+        }
+
+        try model.update(
+            parameters: ModuleParameters.unflattened(sanitizedWeights),
+            verify: .noUnusedKeys
```
lucasnewman: Could this be `.all` instead? It would be nice to catch e.g. shape mismatches and missing keys as well.

Blaizzy: Makes sense, done! ✅

PR #96, opened 2026-03-14
Fill in some missing codecs

lucasnewman:
Adds DAC, Fish S1 DAC, and BigVGAN, which will be used in a follow up to bring over some additional models from python.

### DIFF
--- a/Sources/MLXAudioCodecs/FishS1DAC/FishS1DACTransformer.swift
+++ b/Sources/MLXAudioCodecs/FishS1DAC/FishS1DACTransformer.swift
@@ -1,0 +1,302 @@
+import Foundation
+import MLX
+import MLXFast
+import MLXNN
+
+public final class FishS1ConvNeXtBlock: Module, UnaryLayer {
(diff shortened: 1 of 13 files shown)


Sources/MLXAudioCodecs/FishS1DAC/FishS1DACTransformer.swift
```diff
@@ -0,0 +1,302 @@
...
+
+    public func callAsFunction(_ x: MLXArray, freqsCis: MLXArray?, mask: MLXArray?) -> MLXArray {
+        let batch = x.shape[0]
+        let seqLen = x.shape[1]
+        let kvSize = nLocalHeads * headDim
+
+        let qkv = wqkv(x)
+        var q = qkv[0..., 0..., 0..<kvSize]
+        var k = qkv[0..., 0..., kvSize..<(2 * kvSize)]
+        var v = qkv[0..., 0..., (2 * kvSize)..<(3 * kvSize)]
```
Blaizzy, line 136: We could use MLX split

PR #108, opened 2026-03-21
Fix Parakeet multilingual recognition (Russian/non-English)

### LINKED ISSUE #109: Parakeet TDT v3: Russian output is transliterated garbage — BatchNorm in training mode
## Problem

Parakeet-TDT-0.6B-v3 (`mlx-community/parakeet-tdt-0.6b-v3`) produces transliterated Latin/Polish-diacritic mix instead of Cyrillic when transcribing Russian audio.

**MLX output:**
```
Iscusten intellect miniat. Każdy ten pojaвляются novi modelle для распознавания reci.
```

**NeMo (CUDA, same model, same audio):**
```
Искусственный интеллект меняет мир вокруг нас. Каждый день появляются новые модели для распознавания речи.
```

## Root cause

`BatchNorm` runs in **training mode** during inference. MLX `Module` defaults to `training=true`, so BatchNorm computes statistics from the current batch (size=1) instead of using stored `running_mean`/`running_var`. This produces noisy encoder features that confuse Russian with Polish/Latin transliteration.

Additionally, preprocessing parameters don't match NeMo reference (mel scale, window padding, STFT pad mode, std normalization).

## Fix

PR #108 — the critical fix is one line: `model.train(false)` after loading weights. Plus preprocessing alignment with NeMo.

## Affected models

All Parakeet variants on non-English audio. English works fine because the encoder features are "close enough" even with noisy BatchNorm statistics.

beshkenadze:
## Summary

Fixes #109 — Parakeet TDT v3 produces transliterated Latin/Polish instead of Cyrillic for Russian audio.

**Root cause:** `BatchNorm` running in training mode during inference. MLX `Module` defaults to `training=true`, so BatchNorm computed statistics from the current batch (size=1) instead of using stored `running_mean`/`running_var`. This produced noisy encoder features that confused Russian with Polish/Latin.

**The critical fix is one line:** `model.train(false)` after loading weights.

Also aligns preprocessing with NeMo reference:
- Mel scale: HTK → Slaney
- Window padding: right-pad → center-pad (matching `torch.stft`)
- STFT pad mode: reflect → constant
- Std normalization: ddof=0 → ddof=1 (Bessel's correction)
- Log zero guard: `1e-5` → `2^-24`
- Mel filterbank norm: explicit `"slaney"`
- Special token filtering in decoded output

## Verification

| | Before | After | NeMo (CUDA) |
|---|---|---|---|
| Russian | `Iscusten intellect miniat. Każdy ten...` | `Искусственный интеллект меняет мир...` | `Искусственный интеллект меняет мир...` |
| English | ✅ works | ✅ works | ✅ works |

- Encoder output diff vs NeMo: ~0.001 (was ~0.05)
- Mel spectrogram diff vs NeMo: ~0.005 (normal float precision)

## Test plan

- [x] Russian audio: 100% Cyrillic, matches NeMo reference
- [x] English audio: no regression ("Intention." transcribed correctly)
- [ ] Other languages (German, Spanish, etc.) — needs community testing

### DIFF
--- a/Tests/MLXAudioSTTTests.swift
+++ b/Tests/MLXAudioSTTTests.swift
@@ -1842,7 +1842,22 @@ struct ParakeetSTTTests {
         ])
     }
 
-    @Test func fromDirectorySmokeTestWithFixtureConfigAndWeights() async throws {
+    @Test func tokenizerFiltersSpecialTokens() {
+        let vocab = ["<unk>", "<|nospeech|>", "<pad>", "<|startoftranscript|>",
+                     "<|pnc|>", "<|ru|>", "▁", "h", "e", "l", "o"]
+        let tokens = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 9, 10]
+        let text = ParakeetTokenizer.decode(tokens: tokens, vocabulary: vocab)
+        #expect(text == " hello")
+
+        #expect(ParakeetTokenizer.isSpecialToken(0, vocabulary: vocab) == true)   // <unk>
+        #expect(ParakeetTokenizer.isSpecialToken(2, vocabulary: vocab) == true)   // <pad>
+        #expect(ParakeetTokenizer.isSpecialToken(3, vocabulary: vocab) == true)   // <|startoftranscript|>
+        #expect(ParakeetTokenizer.isSpecialToken(5, vocabulary: vocab) == true)   // <|ru|>
+        #expect(ParakeetTokenizer.isSpecialToken(7, vocabulary: vocab) == false)  // h
+        #expect(ParakeetTokenizer.isSpecialToken(99, vocabulary: vocab) == false) // out of range
+    }
+
+    @Test func modelIsInEvalModeAfterLoading() throws {
         let fixtureDir = FileManager.default.temporaryDirectory
             .appendingPathComponent("parakeet-fixture-\(UUID().uuidString)")
         try FileManager.default.createDirectory(at: fixtureDir, withIntermediateDirectories: true)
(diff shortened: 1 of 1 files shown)


Sources/MLXAudioSTT/Models/Parakeet/ParakeetTokenizer.swift
```diff
@@ -1,12 +1,20 @@
 import Foundation
 
 enum ParakeetTokenizer {
+    static func isSpecialToken(_ tokenId: Int, vocabulary: [String]) -> Bool {
+        guard tokenId >= 0, tokenId < vocabulary.count else { return false }
+        let piece = vocabulary[tokenId]
+        return (piece.hasPrefix("<|") && piece.hasSuffix("|>"))
+            || piece == "<unk>" || piece == "<pad>"
```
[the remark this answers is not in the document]
beshkenadze: Added `tokenizerFiltersSpecialTokens()` test — verifies `<|...|>`, `<unk>`, `<pad>` are excluded from decoded output.

Sources/MLXAudioSTT/Models/Parakeet/ParakeetModel.swift
```diff
@@ -554,6 +558,7 @@ public extension ParakeetModel {
         }
 
         try model.update(parameters: ModuleParameters.unflattened(sanitized), verify: .all)
+        model.train(false)
         eval(model)
         return model
```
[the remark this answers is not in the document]
beshkenadze: Added `model.training == false` assertion to the existing `fromDirectory` smoke test.

Sources/MLXAudioSTT/Models/Parakeet/ParakeetAudio.swift
```diff
@@ -28,27 +28,29 @@ enum ParakeetAudio {
...
+            melScale: .slaney
         )
 
         var mel = MLX.matmul(power, filters.asType(power.dtype))
-        mel = MLX.log(mel + MLXArray(1e-5, dtype: mel.dtype))
+        mel = MLX.log(mel + MLXArray(config.logZeroGuardValue, dtype: mel.dtype))
 
         let normalized: MLXArray
         if config.normalize == "per_feature" {
             let mean = MLX.mean(mel, axis: 0, keepDims: true)
```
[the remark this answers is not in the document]
beshkenadze: Added `melPreprocessingNumericInvariants()` (mean≈0, std≈1, no NaN/Inf) and `melPreprocessingShortAudioDoesNotCrash()` (3-sample edge case) tests.

Sources/MLXAudioSTT/Models/Parakeet/ParakeetModel.swift
```diff
@@ -554,6 +558,7 @@ public extension ParakeetModel {
         }
 
         try model.update(parameters: ModuleParameters.unflattened(sanitized), verify: .all)
+        model.train(false)
```
lucasnewman, line 561: It would be more typical to call `model.eval()` here (as in train vs eval, not mx.eval), which should have the same effect.

beshkenadze: @lucasnewman I got `Value of type 'ParakeetModel' has no member 'eval'` error.

beshkenadze: It seems that the MLX Swift API Module doesn't have an `eval()` method. it's only available in Python MLX (`mlx.nn.Module.eval()`).

lucasnewman: Oh, weird! Ok, it's fine the way you have it then.

PR #125, opened 2026-03-25
[Qwen3 TTS] Cache reference audio between generations

lucasnewman:
Based on the ideas in https://github.com/Blaizzy/mlx-audio-swift/pull/113, but handles concurrent generation properly.


Blaizzy (approved):
LGTM 🚀

PR #126, opened 2026-03-25
Add context support to Qwen3 ASR

### LINKED ISSUE #101: Qwen3 STT support context parameter for better accuracy
Thanks for your excellent work

lucasnewman:
Resolves https://github.com/Blaizzy/mlx-audio-swift/issues/101

### DIFF
--- a/Sources/MLXAudioSTT/Models/Qwen3ASR/Qwen3ASR.swift
+++ b/Sources/MLXAudioSTT/Models/Qwen3ASR/Qwen3ASR.swift
@@ -44,6 +44,7 @@ extension Qwen3ASRModel: STTGenerationModel {
             audio: audio,
             maxTokens: generationParameters.maxTokens,
             temperature: generationParameters.temperature,
+            context: "",
             language: generationParameters.language,
             chunkDuration: generationParameters.chunkDuration,
             minChunkDuration: generationParameters.minChunkDuration
@@ -58,6 +59,7 @@ extension Qwen3ASRModel: STTGenerationModel {
             audio: audio,
             maxTokens: generationParameters.maxTokens,
             temperature: generationParameters.temperature,
+            context: "",
             language: generationParameters.language,
             chunkDuration: generationParameters.chunkDuration,
             minChunkDuration: generationParameters.minChunkDuration
@@ -1138,10 +1140,12 @@ public class Qwen3ASRModel: Module {
         return merged.isEmpty ? nil : merged.joined(separator: ",")
     }
 
-    public func buildPrompt(numAudioTokens: Int, language: String? = nil) -> MLXArray {
-        guard let tokenizer = tokenizer else {
-            fatalError("Tokenizer not loaded")
-        }
+    func buildPromptText(
+        numAudioTokens: Int,
+        context: String = "",
+        language: String? = nil
+    ) -> String {
+        let systemContext = context.isEmpty ? "" : context
 
         let assistantPrefix: String
         if let langName = normalizeLanguageName(language) {
@@ -1150,13 +1154,28 @@ public class Qwen3ASRModel: Module {
             assistantPrefix = ""
         }
 
-        let prompt = "<|im_start|>system\n<|im_end|>\n"
+        return "<|im_start|>system\n\(systemContext)<|im_end|>\n"
             + "<|im_start|>user\n<|audio_start|>"
             + String(repeating: "<|audio_pad|>", count: numAudioTokens)
             + "<|audio_end|><|im_end|>\n"
             + "<|im_start|>assistant\n"
             + assistantPrefix
+    }
 
+    public func buildPrompt(
+        numAudioTokens: Int,
+        context: String = "",
+        language: String? = nil
+    ) -> MLXArray {
+        guard let tokenizer = tokenizer else {
+            fatalError("Tokenizer not loaded")
+        }
+
+        let prompt = buildPromptText(
+            numAudioTokens: numAudioTokens,
+            context: context,
+            language: language
+        )
         let tokenIds = tokenizer.encode(text: prompt)
         return MLXArray(tokenIds.map { Int32($0) }).expandedDimensions(axis: 0)
     }
@@ -1175,6 +1194,7 @@ public class Qwen3ASRModel: Module {
         audio: MLXArray,
         maxTokens: Int,
         temperature: Float,
+        context: String,
         language: String?
     ) -> (text: String, language: String?, promptTokens: Int, generationTokens: Int) {
         guard let tokenizer = tokenizer else {
@@ -1184,7 +1204,11 @@ public class Qwen3ASRModel: Module {
         let eosTokenIds = [151645, 151643]
 
         let (inputFeatures, featureAttentionMask, numAudioTokens) = preprocessAudio(audio)
-        let inputIds = buildPrompt(numAudioTokens: numAudioTokens, language: language)
+        let inputIds = buildPrompt(
+            numAudioTokens: numAudioTokens,
+            context: context,
+            language: language
+        )
         let promptTokenCount = inputIds.dim(1)
 
         let audioFeatures = getAudioFeatures(inputFeatures, featureAttentionMask: featureAttentionMask)
@@ -1238,6 +1262,7 @@ public class Qwen3ASRModel: Module {
         audio: MLXArray,
         maxTokens: Int = 8192,
         temperature: Float = 0.0,
+        context: String = "",
         language: String? = nil,
         chunkDuration: Float = 1200.0,
         minChunkDuration: Float = 1.0
@@ -1269,6 +1294,7 @@ public class Qwen3ASRModel: Module {
                 audio: chunkAudio,
                 maxTokens: remainingTokens,
                 temperature: temperature,
+                context: context,
                 language: forcedLanguage
             )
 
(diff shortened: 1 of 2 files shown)


NicolasArnouts (commented):
## Background & Ecosystem Context

Context/system prompting is a **documented, first-class feature** of Qwen3-ASR. The official model (`QwenLM/Qwen3-ASR`) has a `context` parameter on `transcribe()` designed for **hotword biasing** — providing domain-specific vocabulary (medical terms, product names, proper nouns) to improve recognition accuracy. It supports up to ~10,000 tokens of context.

The prompt format across the ecosystem is consistent:

```
<|im_start|>system\n{context}<|im_end|>\n
<|im_start|>user\n<|audio_start|>{audio_pads}<|audio_end|><|im_end|>\n
<|im_start|>assistant\n{language prefix}
```

### Cross-implementation comparison

| Project | Parameter name | Implementation |
|---------|---------------|----------------|
| **QwenLM/Qwen3-ASR** (official) | `context` | System message content |
| **Blaizzy/mlx-audio** (Python) | `system_prompt` | `_build_prompt()` injects into system turn |
| **k2-fsa/sherpa-onnx** (C++) | hotwords string | Concatenated into system segment |
| **dseditor/QwenASRMiniTool** | `context` | Token-level splice at position 3 |
| **This PR** (Swift) | `context` | String interpolation into system turn |

The PR's naming choice (`context`) aligns with the **official Qwen API**, not the Python mlx-audio's `system_prompt`. This is arguably the better name.

---

## What's Good

1. **Clean refactor**: Splitting `buildPromptText` (returns `String`) from `buildPrompt` (tokenizes to `MLXArray`) is a solid design decision. It makes the text output independently testable without needing a tokenizer.

2. **Complete threading**: Context flows through all 5 call sites — `generate()`, `generateStream()`, `generateSingleChunk()`, `buildPrompt()`, and both `STTGenerationModel` protocol conformance methods.

3. **Concurrency is sound**: `String` is `Sendable`, so context safely crosses the `Task.detached` boundary in the streaming path without needing `UncheckedSendableBox` wrapping.

4. **Each chunk gets the same context**: Both streaming and non-streaming paths pass the same `context` to every chunk's `buildPrompt()` call — correct behavior since context biasing should be consistent across the entire transcription.

5. **Test covers the key behavior**: Verifies context injection, audio token count, and assistant prefix in one focused test.

6. **Consistent with codebase patterns**: GraniteSpeech already has a model-specific `prompt: String?` parameter that's NOT in `STTGenerateParameters` — this PR follows the exact same pattern.

---

## Issues

### 1. Test uses context as natural language instruction — but Qwen3-ASR expects hotword vocabulary

The test uses:
```swift
context: "Prefer product names over pronouns."
```

But Qwen3-ASR's context feature is trained for **hotword biasing** (space-separated vocabulary terms), not natural language instructions. Every official example in `QwenLM/Qwen3-ASR` uses vocabulary terms:
```python
context=["交易 停滞"]  # space-separated hotwords
```

The model treats the system prompt as conditioning tokens for recognition bias, not as an instruction to follow. A natural language sentence like "Prefer product names over pronouns" could cause those words to leak into the transcription output (known issue: QwenLM/Qwen3-ASR#106, #140).

Consider changing the test to use a realistic hotword example:
```swift
context: "TypeWhisper MLX CoreML"
```

This better documents the intended usage for future contributors.

### 2. Trailing newline mismatch with Python reference (minor correctness)

The Python `_build_prompt()` adds a trailing `\n` after the system prompt:

```python
system_content = f"{system_prompt}\n" if system_prompt else ""
```

The PR does **not** add a trailing newline:

```swift
return "<|im_start|>system\n\(context)<|im_end|>\n"
```

With context `"foo bar"`:
- **Python**: `<|im_start|>system\nfoo bar\n<|im_end|>\n`
- **Swift PR**: `<|im_start|>system\nfoo bar<|im_end|>\n`

This may affect tokenization and therefore model behavior. The official Qwen chat template also includes the trailing newline. I'd recommend matching the Python behavior:

```swift
let systemContent = context.isEmpty ? "" : "\(context)\n"
return "<|im_start|>system\n\(systemContent)<|im_end|>\n"
```

### 3. Redundant `systemContext` assignment (nit)

```swift
let systemContext = context.isEmpty ? "" : context
```

This is a no-op — `systemContext` always equals `context`. If the intent was to add the trailing newline (per issue #2 above), this is where it should go. As-is, just use `context` directly.

### 4. `context: ""` hardcoded at protocol boundary — intentional but worth documenting

Both `STTGenerationModel` conformance methods hardcode `context: ""`:

```swift
public func generate(audio: MLXArray, generationParameters: STTGenerateParameters) -> STTOutput {
    generate(
        audio: audio,
        ...
        context: "",  // Always empty through protocol API
        ...
    )
}
```

This means callers using the generic `STTGenerationModel` protocol can never pass context. This is **consistent with how GraniteSpeech handles `prompt`** (also not exposed through the protocol), so it's a valid design choice. However, consider:

- The **Swift CLI already accepts `--context` but prints a warning that it's ignored** (`App.swift:278-280`). This PR doesn't wire it up — is that intentional for a follow-up?
- A one-line doc comment on `context` in `generate()` explaining it's for hotword biasing (space-separated vocabulary terms) would help.

### 5. `buildPromptText` is `internal` while `buildPrompt` is `public` (deliberate?)

`buildPromptText` is `func` (internal) while `buildPrompt` is `public func`. The test accesses it via `@testable import`. If this is intentional (only expose the tokenized version publicly), that's fine. But if external callers would benefit from the text version (e.g., for debugging/logging prompt content), consider making it `public`.

### 6. Missing test cases

The test covers context **with** a value, but doesn't verify:
- **Empty context** (default): that the prompt remains `<|im_start|>system\n<|im_end|>\n` — no stray whitespace
- **No language** (auto-detect): that context + `language=nil` produces the right prompt (no assistant prefix)

### 7. `String = ""` vs `String? = nil` type choice

GraniteSpeech in this same codebase uses `prompt: String? = nil`. The Python mlx-audio uses `system_prompt: str | None = None`. The PR uses `context: String = ""`. Not blocking, but `String? = nil` would be more consistent with existing patterns.

---

## Known Qwen3-ASR Context Gotchas (FYI, not PR issues)

From the Qwen3-ASR issue tracker:
- **Context leaking**: Hotwords can appear verbatim in transcription output (QwenLM/Qwen3-ASR#106)
- **Hotword repetition**: With the 0.6B model in streaming mode, phonetically similar speech can cause the model to repeat the hotword list (QwenLM/Qwen3-ASR#140)
- **Effectiveness is unpredictable**: The Qwen maintainer describes context biasing effectiveness as "玄学" (metaphysical/unpredictable)

---

## Verdict

**Approve with minor changes.** The implementation is solid, correctly threaded through all code paths, concurrency-safe, and follows existing codebase patterns.

**Should fix:**
- Update test to use realistic hotword example instead of natural language instruction (documents correct usage)
- Add trailing `\n` after context to match the Python reference and official Qwen chat template
- Remove the redundant `systemContext` variable

**Nice to have:**
- Add test for empty context (default path)
- Wire up CLI `--context` to actually use this feature (could be a follow-up PR)
- Consider `String? = nil` for consistency with GraniteSpeech's `prompt: String?`
- Doc comment on the `context` parameter explaining it's for space-separated hotword biasing

NicolasArnouts:
Hi @lucasnewman — following up on my review above with a patch PR into your `qwen3asr-context` branch: **lucasnewman/mlx-audio-swift#1**.

It addresses the three should-fix items:

- Appends the trailing `\n` after `context` to match the Python `mlx-audio` reference (`system_content = f"{system_prompt}\n"`) and the official Qwen chat template
- Drops the redundant `systemContext` no-op assignment (folded into the newline logic)
- Adds a doc comment clarifying that `context` is for hotword biasing (space-separated vocabulary terms), not natural-language instructions
- Updates the existing test with a realistic hotword example
- Adds coverage for the empty-context and nil-language paths

If you merge lucasnewman/mlx-audio-swift#1 into `qwen3asr-context`, this PR will pick up the changes automatically. Happy to iterate if you'd like any of it adjusted.

NicolasArnouts:
**Correction to my earlier review** — one of my "should-fix" items was wrong.

I claimed you were missing a trailing `\n` after `context` in the system turn to match Python `Blaizzy/mlx-audio`. After tracing the authoritative sources, **your original format is correct** and Python mlx-audio is the outlier:

| Source | Format |
|---|---|
| HF `chat_template` in `Qwen/Qwen3-ASR-Flash/tokenizer_config.json` | `<\|im_start\|>system\n{content}<\|im_end\|>\n` (no trailing `\n`) |
| `QwenLM/Qwen3-ASR` inference (`apply_chat_template`) | no trailing `\n` |
| `k2-fsa/sherpa-onnx` Qwen3-ASR impl | no trailing `\n` |
| `dseditor/QwenASRMiniTool` (prefix_ids captured from real processor) | no trailing `\n` |
| `Blaizzy/mlx-audio` Python | **trailing `\n`** (hand-rolled, bypasses `apply_chat_template`) |

Tokenizer roundtrip on the Qwen2.5/3 tokenizer also confirms: adding the `\n` inserts token `198` between the content and `<\|im_end\|>` — an extra newline token in a position the model wasn't trained on — which would degrade hotword conditioning rather than fix anything.

So: your prompt format is right. I've updated my patch PR (lucasnewman/mlx-audio-swift#1) to drop that "fix" — it now just contains the non-controversial items:
- Drop redundant `systemContext` ternary
- Doc comment clarifying `context` is for hotword biasing
- Existing test now uses a realistic hotword example
- Add tests for empty-context and nil-language paths

Sorry for the noise on that point. The other items still stand if you want them.

Blaizzy (approved):
LGTM, thanks! 🚀

PR #143, opened 2026-04-04
fix: error: no such module 'Tokenizers' 

christophsturm:


christophsturm:
@lucasnewman my pr contains a test project so if you comment out the fix that should fail.

lucasnewman:
Ok, thanks. It took me a while to hunt this down -- it looks like there was a change in mlx-swift-lm in https://github.com/ml-explore/mlx-swift-lm/pull/118 which triggers this. We're pinned to major version 2 so it shouldn't immediately impact anything unless you've modified the package locally.

christophsturm:
right so my fix improves the situation when using this library together with other versions.

lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-audio-swift/pull/158

PR #147, opened 2026-04-09
feat: add OmniVoice TTS model support

leegang:
Add complete OmniVoice support for mlx-community/OmniVoice-bf16, mirroring the Python omnivoice-infer CLI functionality.

Features:
- Three modes: auto voice, voice design, and voice cloning
- Full CLI integration with all 11 OmniVoice parameters (--instruct, --num_step, --guidance_scale, --speed, --duration, --t_shift, --denoise, --postprocess_output, --layer_penalty_factor, --position_temperature, --class_temperature)
- OmniVoiceGenerateParameters with fast/high-quality presets
- Complete test suite (8 tests passing)
- Comprehensive documentation

Model Architecture:
- Qwen3 LLM backbone (28 layers, 1024 hidden, 16 heads)
- 8 audio codebooks with hierarchical weighting
- 24kHz sample rate output
- 100% parameter compatibility with Python CLI

Files:
- Sources/MLXAudioTTS/Models/OmniVoice/ (model implementation)
- Sources/MLXAudioTTS/TTSModel.swift (factory registration)
- Sources/Tools/mlx-audio-swift-tts/App.swift (CLI integration)
- Tests/OmniVoiceTests.swift (test suite)
- Documentation and testing scripts


lucasnewman (changes requested):
@leegang Thanks for the contribution!

Your change should only include the model itself, a README for the model, and any relevant tests. You shouldn't need to modify the base TTS protocol -- use sensible defaults for the model and specify a custom `generate()` overload if you want to allow additional customization.

Your changes shouldn't include any one-off testing scripts, Markdown files, or other detritus at the root level.

If you can fix those issues up I'm happy to review further.

Sources/Tools/mlx-audio-swift-tts/App.swift
```diff
@@ -50,7 +50,19 @@ enum App {
                 timestamps: args.timestamps,
                 benchmark: args.benchmark,
                 rawIPA: args.rawIPA,
-                language: args.language
+                language: args.language,
+                // OmniVoice-specific parameters
+                instruct: args.instruct,
+                numStep: args.numStep,
```
lucasnewman: Can we just use sensible defaults here? The CLI tool isn't supposed to be a comprehensive test bed with every option for every model -- it's not really tractable. The API surface for the model can support a `generate()` overload with all of these options easily, though, so your own app can use them.

Sources/Tools/mlx-audio-swift-tts/App.swift
```diff
@@ -95,6 +119,43 @@ enum App {
             }
         }
 
+        // Configure OmniVoice-specific parameters if the model is OmniVoice
+        if let omnivoiceModel = loadedModel as? OmniVoiceModel {
+            omnivoiceModel.setGenerationConfig(
```
lucasnewman: This should just be an `generate*()` overload on the model that takes the custom generation config.

lucasnewman:
@leegang Overall this is looking good. I think you should just have a `generate*()` overload that accepts your custom generation parameters structure to make the API as simple as possible. See my comment about the CLI tool changes as well. If we can fix those up we can get this in.

lucasnewman:
Added in https://github.com/Blaizzy/mlx-audio-swift/pull/209

PR #148, opened 2026-04-09
feat: speaker embedding support for Qwen3-TTS CustomVoice models

Ichigo3766:
## Summary

Adds speaker embedding injection for Qwen3-TTS `custom_voice` models (e.g. `mlx-community/Qwen3-TTS-12Hz-0.6B-CustomVoice-8bit`), fixing voice identity not being preserved across sentences.

Also includes a prerequisite fix for flexible JSON decoding of `spkId` and `spkIsDialect` config fields which use mixed types in CustomVoice `config.json` files.

---

## Problem

When using CustomVoice models with a named speaker (e.g. `"ryan"`), the voice was not being applied — the model generated audio with a random/default voice instead of the requested speaker. This is because:

1. **Config decoding failed** — `spkId` values can be `Int` or `[Int]`, and `spkIsDialect` values can be `Bool` or `String`. A plain `Codable` struct would crash or silently drop these fields.

2. **Speaker embedding not injected** — Even with valid config, the `prepareGenerationInputs` path never looked up the speaker ID or injected its embedding into the codec prefix.

---

### `Qwen3TTSModelConfig.swift` — Flexible config decoders
- Added `SpkIdValue` enum (`int` / `array` cases) with `.intValue` helper that returns the first (or only) speaker ID
- Added `SpkDialectValue` enum (`bool` / `dialect` cases) with `.isDialect` and `.dialectName` helpers
- Both enums handle the mixed-type JSON gracefully without schema changes

### `Qwen3TTS.swift` — Speaker embedding injection
- In `generateVoiceDesign()`: detect `ttsModelType == "custom_voice"`, route `voice` as `speaker` (not `instruct`)
- Added `speaker: String?` parameter to `prepareGenerationInputs()`
- Look up `speaker` in `talkerConfig.spkId` → get integer ID → call `talker.getInputEmbeddings()` to produce embedding
- Inject the speaker embedding between the codec think/nothink prefix tokens and the `[codecPad, codecBos]` suffix
- Apply dialect language ID override when `spkIsDialect` indicates a dialect speaker (Cantonese, Hokkien, etc.)

---

## Testing

Tested on device with `mlx-community/Qwen3-TTS-12Hz-0.6B-CustomVoice-8bit` (8-bit quantized):
- Speaker `"ryan"` consistently produces the same voice across multiple sentences ✅
- Speaker `"female_1"` produces a distinctly different voice from `"ryan"` ✅  
- Standard VoiceDesign models (non-CustomVoice) are unaffected — `instruct` path unchanged ✅
- ICL (reference audio) path is unaffected ✅

### DIFF
--- a/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
+++ b/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
@@ -225,7 +225,16 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
             ttsPadEmbed = prepared.2
             refCodes = prepared.3
         } else {
-            let prepared = prepareGenerationInputs(text: text, language: language, instruct: instruct)
+            // For CustomVoice models: voice is a speaker name, not free-text instruct.
+            let isCVModel = config.ttsModelType == "custom_voice"
+            let speaker: String? = isCVModel ? instruct : nil
+            let effectiveInstruct: String? = isCVModel ? nil : instruct
+            let prepared = prepareGenerationInputs(
+                text: text,
+                language: language,
+                instruct: effectiveInstruct,
+                speaker: speaker
+            )
             inputEmbedsInit = prepared.0
             trailingTextHidden = prepared.1
             ttsPadEmbed = prepared.2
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
Thanks!

PR #151, opened 2026-04-13
fix: pass textProcessor through in KokoroModel.fromPretrained

Oliver2213:
fromPretrained accepts a textProcessor parameter but was not forwarding it to fromModelDirectory, so it always defaulted to nil. This meant the G2P pipeline (MisakiTextProcessor / KokoroMultilingualProcessor) never ran — raw English text went directly to the character-level tokenizer, producing garbled speech output.


lucasnewman (approved):
Thanks!

PR #152, opened 2026-04-13
fix: exclude in-tree docs from SwiftPM targets

gaelic-ghost:
Heya! I noticed some of the in-tree docs were getting picked up during build since they weren’t excluded from the package targets, so I added those excludes in `Package.swift`.

Validated locally on Apple Silicon with `xcodebuild` using:
- `MLXAudio-Package`
- `mlx-audio-swift-tts`

Edit: Woops, fixed. Initially, I'd accidentally stacked this atop another branch I'm working on, sorry about that. I need more coffee this morning xD

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -109,7 +109,20 @@ let package = Package(
                 .product(name: "HuggingFace", package: "swift-huggingface"),
                 .product(name: "Transformers", package: "swift-transformers"),
             ],
-            path: "Sources/MLXAudioTTS"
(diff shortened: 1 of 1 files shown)


lucasnewman:
Yeah, it's tedious that swiftpm doesn't support wildcards and we had intentionally left the warnings for the sake of clarity in the Package.swift file before, but I suppose it's fine to just mark these.

lucasnewman (approved):

PR #153, opened 2026-04-13
fix(cohere): load unmodified Python-produced quantized checkpoints

beshkenadze:
## Summary
- add Python-to-Swift checkpoint compatibility for Cohere quantized checkpoints
- load unmodified Python-produced 8-bit and 4-bit artifacts in Swift
- add loader diagnostics and parity tests for fp16, Swift quantized, and Python-produced quantized checkpoints

## Verification
- Swift fp16, Swift 8-bit, Swift 4-bit, Python 8-bit, and Python 4-bit all transcribe the English parity fixture to the same text
- local Cohere runtime harness reproduces matching output on the validated artifacts
- CUDA official path was also checked separately for semantic parity on the English fixture


lucasnewman (approved):
Thanks!

PR #155, opened 2026-04-14
Propagate cancellation to TTS generateStream producers

ashwaryy:
## Problem

Several `Model.generateStream(...)` implementations build an `AsyncThrowingStream` whose producer is an unstructured `Task { @Sendable [weak self] in ... }`, but the `continuation` has no `onTermination` handler. When a consumer cancels (drops the stream or cancels its enclosing task), the producer Task is **not** cancelled, it keeps running until natural completion (`maxTokens` / EOS).

Compounding this, several long-running generation loops never check `Task.isCancelled`, so even if the producer Task were cancelled, the MLX forward-pass loop would still run to the token budget.

In a downstream macOS app this surfaces as: clicking "Stop" mid-generation successfully cancels the Swift task chain, but CPU stays pinned (and KV-cache state continues mutating) until the underlying token loop hits `maxTokens`.

## Fix

Two complementary patches, applied per model:

**A - Wire `onTermination` to cancel the producer Task** in every `generateStream` that spawns an unstructured Task. This matches the canonical pattern already used by `proxyAudioStream` in `Sources/MLXAudioTTS/Generation.swift`:

```swift
let task = Task { @Sendable [weak self] in ... }
continuation.onTermination = { @Sendable _ in task.cancel() }
return stream
```

**B - Add `if Task.isCancelled { break }` at the top of synchronous token loops** that run inside the producer Task, so cancellation actually exits the loop between forward passes.

## Files touched

- `Models/Chatterbox/ChatterboxModel.swift` - A
- `Models/Chatterbox/T3/T3Model.swift` - B
- `Models/Chatterbox/T3/T3GPT2Model.swift` - B
- `Models/FishSpeech/FishSpeechModel.swift` - A + B
- `Models/Llama/LlamaTTS.swift` - A + B (streaming loop already had the check; left untouched)
- `Models/Marvis/MarvisTTSModel.swift` - A on both `generateStream` overloads (loops already had checks)
- `Models/PocketTTS/PocketTTSModel.swift` - A + B
- `Models/Qwen3/Qwen3.swift` - A + B (streaming loop already had the check; left untouched)
- `Models/Qwen3TTS/Qwen3TTS.swift` - A + B
- `Models/Soprano/Soprano.swift` - A + B
- `Models/StyleTTS2/Kokoro/KokoroModel.swift` - A (refactored from inline-closure form to `makeStream()` so the producer Task is bindable; no inner synchronous loop)

## Behavior change

- **Cancellation paths**: producer Tasks now stop within ~1 token of consumer termination. CPU drops to baseline immediately instead of at `maxTokens`.
- **Non-cancellation paths**: no behavior change. `Task.isCancelled` is false in normal completion; `onTermination` only fires when the consumer terminates the stream.

## Verification

- `swift build` clean against the package.
- Verified end-to-end in a downstream macOS app using the Qwen3-TTS model: previously "Stop" left CPU pinned for the remainder of `maxTokens`; with this branch CPU returns to baseline within a single forward pass.

### DIFF
--- a/Sources/MLXAudioTTS/Generation.swift
+++ b/Sources/MLXAudioTTS/Generation.swift
@@ -18,6 +18,16 @@ public protocol SpeechGenerationModel: AnyObject {
         generationParameters: GenerateParameters
     ) async throws -> MLXArray
 
+    func generate(
+        text: String,
+        voice: String?,
+        refAudio: MLXArray?,
+        refText: String?,
+        language: String?,
+        generationParameters: GenerateParameters,
+        seed: UInt64?
+    ) async throws -> MLXArray
+
     func generateStream(
         text: String,
         voice: String?,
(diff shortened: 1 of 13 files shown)


lucasnewman:
@ashwaryy Thanks for the PR! I made a few tweaks on top of your draft and put up a PR with tests for it here: https://github.com/Blaizzy/mlx-audio-swift/pull/157

Issue #149, opened 2026-04-11
Qwen3-TTS: add reusable, consumer-managed reference conditioning for repeat clone generation


gaelic-ghost:
## Summary

Qwen3-TTS recomputes the same clone conditioning on every generation when `refAudio + refText` are reused.

PR #125 approaches this with internal instance-local caching, itself extending PR #113. I have a different solution that exposes reusable conditioning for consumers to manage directly. This is something I’ve prototyped for a package of mine that depends on `mlx-audio-swift`.

I’m curious whether this direction is more widely preferable, which is why I’m opening this issue first. I do have a working branch on my fork [here](https://github.com/gaelic-ghost/mlx-audio-swift/tree/add/qwenReferenceConditioning) for anyone curious.

## Problem

When generating multiple outputs with the same cloned voice, the model repeats the same work each time:

- speaker embedding extraction
- reference audio codec encoding
- reference text preparation
- language metadata resolution
- assembling the request-specific inputs from the above

The current Qwen3 API doesn’t have a clean hook for this, which prevents callers from preparing and reusing precomputed conditioning values intentionally.

## Considerations

A few constraints I noticed mention of, and kept in mind:

- keeping things model-specific, while minding duplication
- keeping the existing surface stable
- avoiding the cache-key / concurrent mutable state problems that came up in #113

## Proposed Solution

- Add a `public struct Qwen3TTSReferenceConditioning` with explicitly public properties and init, so it's serializable for consumers to store and reuse as they wish.
- Split the current clone-prep path out of `prepareICLGenerationInputs(text:refAudio:refText:speakerEmbedding:language:)` into two helpers:
  - `prepareReferenceConditioning(refAudio:refText:speakerEmbedding:language:)` for reusable reference-side conditioning
  - `prepareICLGenerationInputs(text:conditioning:)` for request-specific assembly
- Add overloads to `generate(...)` and `generateStream(...)` that take `conditioning:` instead of raw `refAudio:` and `refText:`
- Keep the current API unchanged, but route it through the same helper split so there is still one source of truth and no duplicated clone logic

lucasnewman:
@gaelic-ghost The ref you posted doesn't seem to be valid/public. Can you just send a PR? I'm open to it.

gaelic-ghost:
> [@gaelic-ghost](https://github.com/gaelic-ghost) The ref you posted doesn't seem to be valid/public. Can you just send a PR? I'm open to it.

[...]

PR #156, opened 2026-04-14
feat: add reusable qwen3 reference conditioning

gaelic-ghost:
## Summary

- add a reusable `Qwen3TTSReferenceConditioning` surface for consumer-managed clone/reference preparation
- split reusable reference-side preparation from request-specific generation input assembly
- add generation overloads that accept `conditioning:`
- keep the existing `refAudio` / `refText` path routed through the same underlying implementation
- add tests covering the reusable conditioning flow

## Why

Presently, when the same `refAudio` and `refText` are reused across multiple generations, the model repeats the same conditioning work each time. This change adds a surface for consumers to do that work once, and receive a type containing the conditioning to store and reuse as they wish, without introducing instance-local cache state.

## Notes

This is the branch I referenced in #149, restored as a clean branch on top of current `main`.

### DIFF
--- a/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
+++ b/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
@@ -58,30 +58,25 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
         language: String?,
         generationParameters: GenerateParameters
     ) async throws -> MLXArray {
-        guard speechTokenizer != nil else {
-            throw AudioGenerationError.modelNotInitialized("Speech tokenizer not loaded")
-        }
-        guard tokenizer != nil else {
-            throw AudioGenerationError.modelNotInitialized("Text tokenizer not loaded")
-        }
+        try requireGenerationComponents()
+        let settings = resolveVoiceDesignGenerationSettings(
+            language: language,
+            generationParameters: generationParameters
+        )
 
-        // VoiceDesign: voice parameter is the instruct (voice description)
-        let instruct = voice
-
-        let audio = generateVoiceDesign(
+        return try generateVoiceDesign(
             text: text,
-            instruct: instruct,
-            language: language ?? "auto",
+            instruct: voice,
+            language: settings.language,
             refAudio: refAudio,
             refText: refText,
-            temperature: generationParameters.temperature,
-            topK: 50,
-            topP: generationParameters.topP,
-            repetitionPenalty: generationParameters.repetitionPenalty ?? 1.05,
-            minP: 0.0,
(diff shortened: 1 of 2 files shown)


Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
```diff
@@ -430,17 +483,135 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
...
+    private func requireGenerationComponents() throws {
+        guard speechTokenizer != nil else {
+            throw AudioGenerationError.modelNotInitialized("Speech tokenizer not loaded")
+        }
+        guard tokenizer != nil else {
+            throw AudioGenerationError.modelNotInitialized("Text tokenizer not loaded")
+        }
+    }
+
+    private func resolveVoiceDesignGenerationSettings(
```
lucasnewman, line 622: I'm not sure I get the idea of this struct -- it's seems the same as GenerateParameters with the language from the conditioning, but we already have the default generate params to resolve the sampling defaults, so it seems like we're getting a net code / complexity increase by introducing this additional type at the call sites.

lucasnewman:
@gaelic-ghost Can you resolve the conflicts with the main branch and see the comment?

Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
gaelic-ghost: So, `GenerateParameters` models some basic generation parameters for inference. I'm working here with the same datatypes that your PR #125 was, with the addition of the computed token ids from `refText:`.

As for the API, the overloads seemed like a cleaner, more concentrated option than an 2-case enum w/ associated values. The latter would require changes to the `SpeechGenerationModel` protocol, probably by way of a couple constrained extensions.

And the benefits of this approach vs the post-PR #125 situation include:
 - Persistent and consistent cloned voices
 - Persistent storage of more than one durable voice
 - Being able to switch between voices without regenerating conditioning and introducing drift
 - Reliable generation of cloned voices
 - Avoiding a ton of "bad generation" and "voice drift" edge cases that remain after, or were introduced by, the merge of #125 (yes, i have tested these, and have a patch for some of the new ones, though that's a whole 'nother can of worms)

As a use-case example, using two cloned voices for the popular podcast-style format would still require regenerating conditioning on every request without a solution like this, and voices continue to drift a bit back and forth on every request.  

I could slim the struct down by one field, if that would help. And I could cut down on the loc changes by doing this in a messier, harder-to-follow way.

And I'm of course happy to resolve the merge conflicts after I adapt this around the newly merged caching setup.

lucasnewman (approved):
LGTM

PR #157, opened 2026-04-16
Improved task cancellation handling for generation

### LINKED ISSUE #154: Propagate cancellation to TTS generateStream producers
Several `Model.generateStream(...)` implementations build an `AsyncThrowingStream` whose producer is an unstructured `Task { @Sendable [weak self] in ... }`, but the `continuation` has no `onTermination` handler. When a consumer cancels (drops the stream or cancels its enclosing task), the producer Task is **not** cancelled, it keeps running until natural completion (`maxTokens` / EOS).

Compounding this, several long-running generation loops never check `Task.isCancelled`, so even if the producer Task were cancelled, the MLX forward-pass loop would still run to the token budget.

In a downstream apps this surfaces as: clicking "Stop" mid-generation successfully cancels the Swift task chain, but CPU stays pinned (and KV-cache state continues mutating) until the underlying token loop hits `maxTokens`.

lucasnewman:
Built on https://github.com/Blaizzy/mlx-audio-swift/pull/155 -- this allows cancellation generation tasks at the app level and existing the generation loops as quickly as possible across all models. Resolves https://github.com/Blaizzy/mlx-audio-swift/issues/154


Blaizzy (approved):
🚀

PR #158, opened 2026-04-17
Upgrade package versions and adopt changes from mlx-lm-swift

lucasnewman:
There were a number of changes upstream and a new `Tokenizer` class was defined in mlx-lm-swift that conflicts with the swift-huggingface implementation. This upgrades the packages and fixes all the conflicts, so developers can use the latest versions of those packages. See further discussion in https://github.com/Blaizzy/mlx-audio-swift/pull/143


Blaizzy (approved):
LGTM! 🚀

PR #161, opened 2026-04-18
models: escape underscore member reads in EnglishG2P

### LINKED ISSUE #160: EnglishG2P.swift: escape underscore-member reads
## Summary

`EnglishG2P.swift` still has a few bare reads against the underscore-named member on `MToken`, such as:

```swift
$0._.stress
token._.prespace
tokens.first?._.is_head
```

That should be normalized to the escaped form already used elsewhere in the same file:

```swift
$0.`_`.stress
token.`_`.prespace
tokens.first?.`_`.is_head
```

## Why

This caused a plain SwiftPM compile failure downstream for me. Escaping the remaining read sites fixed it.

## File

```text
Sources/MLXAudioTTS/Models/StyleTTS2/G2P/EnglishG2P.swift
```

## Fix

Replace the remaining bare underscore-member reads with escaped reads. This is a small syntax-only fix and does not change behavior.

gaelic-ghost:
## Summary

- escape the remaining underscore-member reads in `EnglishG2P.swift`
- make those reads match the escaped underscore-member writes already used elsewhere in the file
- keep the change syntax-only with no behavior change

## Verification

- `swiftc -frontend -parse Sources/MLXAudioTTS/Models/StyleTTS2/G2P/EnglishG2P.swift`

Closes #160.

### DIFF
--- a/Sources/MLXAudioTTS/Models/StyleTTS2/G2P/EnglishG2P.swift
+++ b/Sources/MLXAudioTTS/Models/StyleTTS2/G2P/EnglishG2P.swift
@@ -243,15 +243,15 @@ final class EnglishG2P {
   }
   
   func mergeTokens(_ tokens: [MToken], unk: String? = nil) -> MToken {
-    let stressSet = Set(tokens.compactMap { $0._.stress })
-    let currencySet = Set(tokens.compactMap { $0._.currency })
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
LGTM

PR #163, opened 2026-04-18
models: hide marvis rope caches from strict verification

### LINKED ISSUE #162: Marvis strict verification fails under newer MLX stack
## Summary

Plain upstream Marvis source can fail strict parameter verification under newer compatible MLX dependency versions because `CSMLlama3ScaledRoPE` exposes runtime-only cache fields as reflected parameter keys.

## Affected Surface

- public API: `TTS.loadModel(modelRepo:)`
- model repo: `Marvis-AI/marvis-tts-250m-v0.2-MLX-8bit`
- source baseline: plain `upstream/main`
- reproduced when the package resolves with:
  - `mlx-swift` `0.31.3`
  - `mlx-swift-lm` `2.31.3`

## Reproduction

Using plain upstream Marvis source on `upstream/main`, reproduce under the newer compatible MLX versions that downstream consumers can normally resolve:

- `mlx-swift` `0.31.3`
- `mlx-swift-lm` `2.31.3`

Then call the normal public Marvis loading API twice concurrently:

```swift
await withThrowingTaskGroup(of: Void.self) { group in
    for _ in 0 ..< 2 {
        group.addTask {
            _ = try await TTS.loadModel(
                modelRepo: "Marvis-AI/marvis-tts-250m-v0.2-MLX-8bit"
            )
        }
    }
    try await group.waitForAll()
}
```

Observed failure:

```text
keyNotFound(path: ["model", "decoder", "layers", "0", "self_attn", "rope", "sinF32"], modules: ["MarvisTTSModel", "CSMModel", "CSMLlamaModel", "CSMTransformerBlock", "CSMLlamaAttention", "CSMLlama3ScaledRoPE"])
```

## Expected Behavior

Strict verification should validate checkpoint-backed model weights only. Runtime RoPE caches should not appear in the reflected parameter surface and should not be demanded from the checkpoint.

## Root Cause

`CSMLlama3ScaledRoPE` stores these runtime-only caches as ordinary non-underscored fields:

- `cosF32`
- `sinF32`
- `cosByDType`
- `sinByDType`

`mlx-swift` module reflection treats those as valid parameter keys unless they are hidden. Under the newer `mlx-swift` line, `Module` builds its reflection cache lazily instead of during `super.init()`, so those caches are visible by the time strict verification runs.

## Fix

Rename the runtime-only RoPE cache fields to underscore-prefixed stored properties:

- `_cosF32`
- `_sinF32`
- `_cosByDType`
- `_sinByDType`

That keeps strict verification on for real checkpoint weights while keeping derived runtime caches out of the reflected parameter set.

## Verification

With the exact same newer compatible MLX dependency versions and the exact same concurrent public `TTS.loadModel(...)` repro:

- plain upstream source fails
- applying only the underscore-prefixed RoPE cache fix makes both concurrent loads pass

gaelic-ghost:
## Summary

- rename the runtime-only Marvis RoPE cache fields to underscore-prefixed stored properties
- keep those derived caches out of MLX module parameter reflection during strict verification
- leave model behavior unchanged

## Why

Under the newer MLX stack, `CSMLlama3ScaledRoPE` can expose runtime cache fields like `sinF32` as reflected parameter keys. That makes strict verification treat them as if they were checkpoint-backed weights.

The fix is to keep those runtime-only cache fields underscore-prefixed:

- `_cosF32`
- `_sinF32`
- `_cosByDType`
- `_sinByDType`

## Verification

- `swiftc -frontend -parse Sources/MLXAudioTTS/Models/Marvis/CSMLlamaModel.swift`

Closes #162.

### DIFF
--- a/Sources/MLXAudioTTS/Models/Marvis/CSMLlamaModel.swift
+++ b/Sources/MLXAudioTTS/Models/Marvis/CSMLlamaModel.swift
@@ -14,10 +14,12 @@ final class CSMLlama3ScaledRoPE: Module {
     let highFreqFactor: Float
     let oldContextLen: Float
 
-    private var cosF32: MLXArray?
-    private var sinF32: MLXArray?
-    private var cosByDType: [DType: MLXArray] = [:]
-    private var sinByDType: [DType: MLXArray] = [:]
+    // Keep runtime-only RoPE caches underscore-prefixed so MLX Module reflection
+    // does not treat them as checkpoint-backed parameters during strict verify.
+    private var _cosF32: MLXArray?
+    private var _sinF32: MLXArray?
+    private var _cosByDType: [DType: MLXArray] = [:]
+    private var _sinByDType: [DType: MLXArray] = [:]
     private var isCacheBuilt = false
 
     init(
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
Thanks!

PR #165, opened 2026-04-22
feat(Qwen3TTS): multi-voice reference audio cache with voiceKey support

Oliver2213:
Extend the single-entry ObjectIdentifier-based ref audio cache to a per-voice dictionary keyed by caller-provided voiceKey string. This avoids cache thrashing when switching between cloned voices (e.g. narrator vs character) on the same model instance.

- ReferenceAudioContext now caches refTextIds and refText alongside speakerEmbedding, refCodes, and codecEmbedIcl
- Add generateStream overload with explicit topK and voiceKey params
- Add clearRefCache() and clearRefCache(forVoice:) public API
- Make speechTokenizer and Qwen3TTSSpeechTokenizer public
- Make hasEncoder public for ICL support checks

Co-written with Claude Opus 4.6.

### DIFF
--- a/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
+++ b/Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
@@ -12,18 +12,37 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
     let config: Qwen3TTSModelConfig
     let talker: Qwen3TTSTalkerForConditionalGeneration
     var speakerEncoder: Qwen3TTSSpeakerEncoder?
-    var speechTokenizer: Qwen3TTSSpeechTokenizer?
+    public var speechTokenizer: Qwen3TTSSpeechTokenizer?
     var tokenizer: Tokenizers.Tokenizer?
 
+    // MARK: - Multi-voice reference audio cache
+
     private struct ReferenceAudioContext {
-        let audioID: ObjectIdentifier
         let speakerEmbedding: MLXArray?
         let refCodes: MLXArray
+        let refTextIds: MLXArray
+        let refText: String
         let codecEmbedIcl: MLXArray
     }
 
-    private let inputPreparationCacheLock = NSLock()
-    private var cachedReferenceAudioContext: ReferenceAudioContext?
+    private let cacheLock = NSLock()
+    private var voiceContextCache: [String: ReferenceAudioContext] = [:]
+
+    private func withCacheLock<T>(_ body: () -> T) -> T {
+        cacheLock.lock()
+        defer { cacheLock.unlock() }
+        return body()
+    }
+
+    /// Clears the reference audio cache for a specific voice.
+    public func clearRefCache(forVoice voiceKey: String) {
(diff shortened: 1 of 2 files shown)


Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
```diff
@@ -111,6 +130,33 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
...
+    /// so this overload lets callers pass it directly.
+    /// voiceKey identifies the voice for multi-voice cache lookup.
+    public func generateStream(
+        text: String,
+        voice: String?,
+        refAudio: MLXArray?,
+        refText: String?,
+        language: String?,
+        generationParameters: GenerateParameters,
+        topK: Int,
```
lucasnewman: I think it would be better to just have top-k in GenerateParameters as an optional int -- what do you think?

Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
```diff
@@ -222,41 +263,51 @@ public final class Qwen3TTSModel: Module, SpeechGenerationModel, @unchecked Send
...
         }
 
         let speakerEmbedding = extractSpeakerEmbedding(refAudio)
         let refCodes = speechTokenizer.encode(referenceAudioForEncoder(refAudio))
         let codecEmbedIcl = codecEmbedIcl(from: refCodes, talkerConfig: talkerConfig)
+
+        guard let tokenizer else {
+            fatalError("Tokenizer not loaded")
+        }
+        let refChatText = "<|im_start|>assistant\n\(refText)<|im_end|>\n"
```
lucasnewman, line 265: Have you measured the performance of this? I would expect it to be so fast it's in the noise to tokenize the text, which adds to the caching complexity here.

lucasnewman:
@Oliver2213 I think this overall looks reasonable, but see what you think about my comments.

Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
Oliver2213: Yeah, cleaner that way / with the other params.

Sources/MLXAudioTTS/Models/Qwen3TTS/Qwen3TTS.swift
Oliver2213: I have not; I just saw it being reran and had it cached since ref text doesn't change for a cloned voice / tokens shouldn't either.

Oliver2213:
@lucasnewman how's this.

lucasnewman:
@Oliver2213 Can you try the changes in https://github.com/Blaizzy/mlx-audio-swift/pull/156 (which is a bit more general) and then see if you still need changes? If so, let me know and I can reopen.

Oliver2213:
Yeah #156 solves the problem for me; I can manage the Qwen3TTSReferenceConditioning objects myself. One minor thing though, the topK on GenerationParameters change - I'll probably PR this separately.

PR #166, opened 2026-04-23
Add optional progressHandler to ModelUtils.resolveOrDownloadModel

Oliver2213:
- Add a `progressHandler` parameter (`(@MainActor @Sendable (Progress) -> Void)?`) to `resolveOrDownloadModel`, defaulting to nil
- When provided, forwards it to `HubClient.snapshot`; otherwise falls back to the existing print-based handler

This lets callers provide their own closure to handle download progress updates (in UI).


lucasnewman (approved):
LGTM

PR #168, opened 2026-04-25
Fix (Qwen3TTS): pass topK and minP from GenerateParameters

Oliver2213:
Use `generationParameters.topK` and `generationParameters.minP` instead of hardcoded values (50 and 0.0) in `resolveVoiceDesignGenerationSettings`


lucasnewman (approved):
Thanks!

PR #169, opened 2026-04-25
Add MOSS-TTS-Nano model

lucasnewman:
A couple of other changes outside the scope of the model itself:
1) The SentencePiece tokenizer needs to handle BPE models in addition to unigram, so I renamed the tokenizer to be more clear about what it now supports.
2) This is the first model that emits multi-channel audio, so I added support for that in the cli tool.


Blaizzy (approved):
LGTM! 🚀

PR #172, opened 2026-04-28
Fix Qwen3ASR stream cancellation propagation

summereasy:
## Summary

PR #157 fixed task cancellation for TTS stream producers. This PR applies the same cancellation bridge to Qwen3ASR.

Qwen3ASR already calls `Task.checkCancellation()` in its chunk and token loops, but its detached producer task was not cancelled when the `AsyncThrowingStream` consumer stopped iterating. This change stores the detached task and cancels it from `continuation.onTermination`.

Other ASR models are left unchanged because their stream implementations do not follow this detached-producer shape and need a broader cancellation pass.

## Verification

```bash
xcodebuild -scheme MLXAudioSTT -destination 'platform=macOS' build
```

Also manually verified in a downstream cancellation flow.


lucasnewman (approved):
Thanks!

PR #179, opened 2026-05-06
Add MOSS-TTS model family

### LINKED ISSUE #171: support MOSS
https://huggingface.co/OpenMOSS-Team

lucasnewman:
This adds the MOSS-TTS/TTSD models, ported from the python implementation.

I also did some mild refactoring of the TTS model loading to make it easier to use path-based models, e.g. embedded in the app bundle on iOS, which was previously not really possible with the generic model loading entrypoints.

Resolves https://github.com/Blaizzy/mlx-audio-swift/issues/171


Blaizzy (approved):
🚀

PR #195, opened 2026-06-05
Add Nemotron 3.5 ASR streaming model support

vanch007:
## Summary
- Add `NemotronASRModel` for mlx-community Nemotron 3.5 ASR streaming checkpoints
- Implement Nemotron-specific config parsing, causal FastConformer, chunked-limited attention mask, language prompt kernel, tokenizer, and greedy RNNT decoding
- Wire Nemotron repos into the STT CLI and README model table

## Supported checkpoints
- `mlx-community/nemotron-3.5-asr-streaming-0.6b`
- `mlx-community/nemotron-3.5-asr-streaming-0.6b-8bit`

## Tests
- `swift test --filter NemotronASRTests`

Note: MLX runtime shape/decode tests are gated behind `MLXAUDIO_ENABLE_MLX_RUNTIME_TESTS=1` because `swift test` on this machine does not package the MLX `default.metallib`; default Nemotron tests still cover config parsing and tokenizer behavior.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -146,6 +146,7 @@ let package = Package(
                 "Models/FireRedASR2/README.md",
                 "Models/GLMASR/README.md",
                 "Models/GraniteSpeech/README.md",
+                "Models/NemotronASR/README.md",
                 "Models/Parakeet/README.md",
                 "Models/Qwen3ASR/README.md",
                 "Models/SenseVoice/README.md",
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ MLXAudio follows a modular design allowing you to import only what you need:
 - **MLXAudioCore**: Base types, protocols, and utilities
 - **MLXAudioCodecs**: Audio codec implementations (SNAC, Encodec, Vocos, Mimi, DACVAE)
 - **MLXAudioTTS**: Text-to-Speech models (Qwen3-TTS, Fish Audio S2 Pro, Soprano, VyvoTTS, Orpheus, Marvis TTS, Pocket TTS)
-- **MLXAudioSTT**: Speech-to-Text models (Qwen3-ASR, Voxtral Realtime, Cohere Transcribe, Parakeet, GLMASR)
+- **MLXAudioSTT**: Speech-to-Text models (Qwen3-ASR, Voxtral Realtime, Cohere Transcribe, Parakeet, Nemotron ASR, GLMASR)
 - **MLXAudioVAD**: Voice Activity Detection & Speaker Diarization (Sortformer, SmartTurn)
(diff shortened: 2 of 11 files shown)


beshkenadze:
Tried this PR against a known-good reference (NeMo CUDA / the Python `nemotron_asr`) and hit a **blocking runtime bug** — the model crashes at weight load before any inference:

```
Error: incompatibleItems(path: ["prompt_kernel"], modules: ["NemotronASRModel"],
  module: PromptKernel { 0: Linear(1152→2048), 2: Linear(2048→1024) }
  value:  [ [bias,weight], none, [bias,weight] ])
```

**Cause:** `NemotronASRPromptKernel` declares its children with integer `@ModuleInfo` keys `"0"` / `"2"`. MLX-swift interprets all-integer child keys as an **array**, so the checkpoint's `prompt_kernel.0` / `prompt_kernel.2` load as `[item0, none(1), item2]` (gap at index 1), which can't match the module → `incompatibleItems`. This isn't caught by CI because the MLX-runtime tests are gated off (`MLXAUDIO_ENABLE_MLX_RUNTIME_TESTS`, the `default.metallib` packaging issue), so only config/tokenizer tests run.

**Fix (2 lines):** give the children non-integer keys and remap the checkpoint keys.

```swift
// NemotronASRPromptKernel
@ModuleInfo(key: "linear0") var linear0: Linear
@ModuleInfo(key: "linear2") var linear2: Linear
```
```swift
// remapKey(_:)
newKey = newKey.replacingOccurrences(of: "prompt_kernel.0.", with: "prompt_kernel.linear0.")
newKey = newKey.replacingOccurrences(of: "prompt_kernel.2.", with: "prompt_kernel.linear2.")
```

**After the fix it runs and is parity-faithful** (built `mlx-audio-swift-stt`, ran on `mlx-community/nemotron-3.5-asr-streaming-0.6b`):
- single clip → `"Hello World, this is a test."` — token-exact with the NeMo/Python reference
- FLEURS en-US (200 utt, 16 kHz, default `att_context=[56,13]`): **WER 9.62% (410/4261)** vs NeMo **CUDA** 9.58% (408) → **+0.04% (2 words)**. Faithful to the CUDA reference at scale.

(Runtime note for repro: copying `mlx-swift_Cmlx.bundle` next to the built CLI works around the metallib-not-packaged issue.)

One follow-up beyond this PR: it's **offline-only** (the `generateStream` here is buffered re-decode); the model's native **cache-aware streaming** isn't implemented yet. Happy to contribute that on top once this lands. Nice work overall — the preprocessing/conformer/decoding are correct.

lucasnewman (approved):
🚀

PR #196, opened 2026-06-05
Nemotron 3.5 ASR: fix prompt_kernel load crash + add cache-aware streaming (builds on #195)

beshkenadze:
## Summary

Builds on @vanch007's #195 (Nemotron 3.5 ASR) and adds the two things needed to make it
production-ready: a **fix for a blocking weight-load crash** and **cache-aware streaming**
(the model's native low-latency mode). Validated against the NeMo **CUDA** reference.

> This branch includes #195's commits plus the fix + streaming, so it supersedes #195.
> Full credit to @vanch007 for the original port — happy to instead submit just the
> streaming on top once #195 lands, whichever the maintainers prefer.

## 1. Fix: `prompt_kernel` load crash (blocking)

#195 crashes at weight load before any inference:
```
incompatibleItems(path: ["prompt_kernel"], ... value: [ [bias,weight], none, [bias,weight] ])
```
Cause: `NemotronASRPromptKernel` used integer `@ModuleInfo` keys `"0"`/`"2"`, which MLX-swift
treats as an array (gap at index 1). Not caught by CI because the MLX-runtime tests are gated
off (the `default.metallib` packaging issue), so only config/tokenizer tests run. Fix: remap to
`linear0`/`linear2` keys.

## 2. Feature: cache-aware streaming

#195's `generateStream` is buffered re-decode (overlapping windows). This adds the model's
**native cache-aware streaming** (`NemotronASRStreaming.swift`): per-layer attention
(left-context) + causal-conv caches, plus incremental causal subsampling (16-frame mel cache).
O(n), no recompute; `generateStream` is now genuinely streaming.

## Validation (FLEURS en-US, 200 utt, 16 kHz; word-level WER, shared normalization)

| Path | att_ctx | WER | Δ vs NeMo CUDA |
|------|---------|-----|----------------|
| NeMo (reference, CUDA RTX 4090) | `[56,13]` | 9.58% (408/4261) | — |
| #195 offline (after the fix) | `[56,13]` | 9.62% (410) | +0.04% |
| **cache-aware streaming (this PR)** | `[56,13]` | **9.43% (402)** | −0.15% |

Single clip is token-exact with the reference (`"Hello World, this is a test."`). The small
offline↔streaming transcript diffs are bf16 rounding (chunked vs full-sequence accumulation),
not a logic bug — WER is equivalent.

### DIFF
--- a/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
+++ b/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
@@ -136,71 +136,83 @@ public final class NemotronASRModel: Module, STTGenerationModel {
         AsyncThrowingStream { continuation in
             let audio1D = self.normalizeAudioToMono(audio).asType(.float32)
             let sampleRate = self.preprocessConfig.sampleRate
-            let totalSamples = audio1D.shape[0]
-            let audioDuration = Double(totalSamples) / Double(sampleRate)
-            let requestedChunk = Double(generationParameters.chunkDuration)
-            let chunkDuration = requestedChunk >= 1199 ? 5.0 : max(0.5, requestedChunk)
-            let overlapDuration = 1.0
+            let audioDuration = Double(audio1D.shape[0]) / Double(sampleRate)
+            let mel = NemotronASRAudio.logMelSpectrogram(audio1D, config: self.preprocessConfig)
+            let frameSeconds = Double(self.encoderConfig.subsamplingFactor * self.preprocessConfig.hopLength)
+                / Double(sampleRate)
 
-            let chunkSamples = max(1, Int(chunkDuration * Double(sampleRate)))
-            let overlapSamples = max(0, min(chunkSamples - 1, Int(overlapDuration * Double(sampleRate))))
-            let stepSamples = max(1, chunkSamples - overlapSamples)
-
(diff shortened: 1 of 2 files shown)


beshkenadze:
Marking draft: this is **stacked on #195**, so the diff currently includes #195's commits — including its `ParakeetConfig.swift` change (making Parakeet RNN-T config inits `public` so NemotronASR can reuse them) and an unsigned commit. Those are #195's, not part of the streaming work (my commit only touches `NemotronASR/`). Once #195 merges I'll rebase this to a clean, streaming-only, signed diff with no Parakeet blast. The load-bug fix is already noted on #195.

lucasnewman:
@beshkenadze Ok, I'll merge in a moment and then we can merge yours.

lucasnewman:
@beshkenadze Can you fix the merge conflict?

beshkenadze:
Rebased onto `main` now that #195 is merged. The diff is clean: only `NemotronASR/` (2 files, +180/−49), no `ParakeetConfig` blast, single signed commit. Builds against current `main`. Still includes the prompt_kernel load-crash fix (`@ModuleInfo(key: "0"/"2")` → `"linear0"/"linear2"` + checkpoint remap) since the merged `main` still has the integer keys that crash at load. Ready to merge.

lucasnewman (approved):
LGTM, thanks!

PR #199, opened 2026-06-08
[Perf] Run the Parakeet Conformer encoder on the Apple Neural Engine (CoreML)

beshkenadze:
# Run the Parakeet Conformer encoder on the Apple Neural Engine (CoreML)

## Summary

Adds an **optional CoreML/ANE path for the Parakeet Conformer encoder**, behind a flag.
The encoder (≈95% of the compute) runs on the Apple Neural Engine via CoreML while the
TDT decoder and chunking stay in MLX. Same transcript, lower power, and a small speedup
on top.

It plugs into the existing `EncoderExecutionImplementation` hook in `ParakeetModel`
(new `.coreML` case + `enableCoreMLEncoder(modelURL:)`), so the decode path is untouched.
It falls back to the MLX encoder if CoreML is unavailable. **Public `MLModel` +
`MLComputeUnits` only — no private ANE APIs.**

```bash
# auto-downloads the prebuilt encoder from Hugging Face
mlx-audio-swift-stt --model beshkenadze/parakeet-tdt-0.6b-v3-mlx-fp16 \
    --audio in.wav --output-path out --ane --chunk-duration 9.95
```

```swift
// .off (default) · .on (default HF repo) · .repo("id") · .package(localURL)
let model = try await ParakeetModel.fromPretrained(repo, aneEncoder: .on)
```

## Why

ANE has no public API — CoreML is the only sanctioned route, and MLX is GPU/Metal only.
Splitting the graph at the encoder boundary (static feed-forward → CoreML/ANE; the
autoregressive TDT loop → MLX) is a clean, lossless way to reach the ANE. The payoff is
mostly **power/thermal** (the encoder leaves the GPU), with a speedup as a bonus.

## Results

Measured on **M1 Max**, `parakeet-tdt-0.6b-v3`, a 20.8-min TED-LIUM 3 talk, chunk 9.95s.

| Metric | all-MLX | hybrid (CoreML/ANE encoder) |
|---|---|---|
| **ANE residency** (encoder) | — | **100%** (0 CPU / 0 GPU ops, 0 graph interruptions) |
| **WER** vs reference | 7.28% | **7.11%** · agreement **1.07%** |
| **RTF** (Swift release, interleaved) | ~95× | **~131×** → **~1.38×** |
| **GPU power** (sustained) | 17.3 W | **3.0 W** (**÷5.8**) |
| **Package power** | 23.4 W | **10.3 W** (**÷2.3**) — ANE encoder ≈ 0.9 W |

- The transcript is reproduced ~1:1 (2786 vs 2802 words); the residual is the
  fp16-vs-bf16 difference (CoreML-fp16 is actually *closer* to fp32 than the shipped
  MLX-bf16 encoder).
- Residency **holds at the 1.1b** variant too (100%, 0 interruptions).

## How it works

- **Fixed input shape** (a fixed mel-frame count, e.g. 1000 = 10s) — required for ANE
  residency; a dynamic (`RangeDim`) time axis drops it to 0%. The Swift wrapper pads each
  chunk's mel to the fixed length and crops the output via the subsampling formula. Keep
  `--chunk-duration ≤ frames·10ms`.
- The ANE output `MLMultiArray` is stride-padded, so the wrapper reads it by strides.
- Conversion (NeMo → `.mlpackage`) lives in `tools/coreml-ane/` (`convert_encoder.py` +
  `convert_traced.py`); see the README. `--fp16-io` gives 100% ANE / 0 CPU ops.

## Scope

- Swift: `ParakeetCoreMLEncoder.swift`, `ParakeetModel.swift` (the `.coreML` case),
  `App.swift` (the `--coreml-encoder` flag).
- Tooling: `tools/coreml-ane/` converter + README.

## Limitations / follow-ups

- The `.mlpackage` is not bundled (it's large). A prebuilt one is hosted on Hugging Face
  ([`beshkenadze/parakeet-tdt-0.6b-v3-coreml-ane`](https://huggingface.co/beshkenadze/parakeet-tdt-0.6b-v3-coreml-ane));
  users can also convert it via the tooling.
- MLX↔CoreML marshaling currently uses CPU copies; a zero-copy IOSurface-backed
  `MLMultiArray` would lift the Swift RTF further (the power win is independent).
- RTF numbers are M1 Max; newer ANE generations should do better.

## Testing

- New CI-safe unit tests (`Tests/ParakeetCoreMLEncoderTests.swift`, swift-testing): the
  output-length math matches the dw-striding formula, and a missing `.mlpackage` throws
  (→ MLX fallback). No ANE/model/network needed; `swift test`: 2/2 pass.
- Builds clean (`release`); transcript parity verified against the all-MLX path on the
  full talk.
- The decode path is unchanged (the encoder is swapped behind the existing hook).

### DIFF
--- a/Sources/MLXAudioSTT/Models/Parakeet/ParakeetCoreMLEncoder.swift
+++ b/Sources/MLXAudioSTT/Models/Parakeet/ParakeetCoreMLEncoder.swift
@@ -1,0 +1,142 @@
+#if canImport(CoreML)
+import CoreML
+import Foundation
+import MLX
+
+/// Drop-in CoreML/ANE replacement for the MLX Conformer encoder. The model is
+/// fixed-shape because ANE requires it (RangeDim → 0% residency), so each chunk's mel
+/// is padded to `fixedFrames` and the output cropped back to the true subsampled length.
+public final class ParakeetCoreMLEncoder: @unchecked Sendable {
+    private let model: MLModel
+    private let featIn: Int
+    /// The fixed mel-frame count the model was converted at (e.g. 1000 = 10 s). Callers use
+    /// this to keep their chunking ≤ this length, since longer mel is cropped.
+    public let fixedFrames: Int
+    private let subsamplingFactor: Int
+    private let inputName: String
+    private let outputName: String
+    /// Expected encoder output width (`d_model`); when set, `encode` rejects a model whose
+    /// output doesn't match the loaded checkpoint instead of returning mis-shaped features.
+    private let expectedDModel: Int?
+    /// One subsampling stage on a length. Parakeet: `(L-1)/2+1`; encoders with different conv
+    /// padding (e.g. Nemotron's pad-before-stride `floor(L/2)+1`) inject their own so the ANE
+    /// output is cropped to the right valid width.
+    private let subsampledLengthStep: (Int) -> Int
+
+    public init(
+        modelURL: URL,
+        featIn: Int,
+        fixedFrames: Int,
+        subsamplingFactor: Int,
+        dModel: Int? = nil,
+        subsampledLengthStep: (@Sendable (Int) -> Int)? = nil,
+        computeUnits: MLComputeUnits = .all,
+        inputName: String = "features",
+        outputName: String = "encoded"
+    ) throws {
+        let compiledURL: URL
+        if modelURL.pathExtension == "mlmodelc" {
+            compiledURL = modelURL
+        } else {
+            compiledURL = try MLModel.compileModel(at: modelURL)
+        }
+        let config = MLModelConfiguration()
+        config.computeUnits = computeUnits
+        self.model = try MLModel(contentsOf: compiledURL, configuration: config)
+        self.featIn = featIn
+        self.fixedFrames = fixedFrames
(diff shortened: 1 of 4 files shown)


lucasnewman:
@beshkenadze It seems like the CoreML export should live in the python repo? Having python tools in this repo doesn't make a lot of sense to me as the toolchains are very distinct. Thoughts?

beshkenadze:
Agreed. I'll move `tools/coreml-ane/` out and co-locate each converter with its `.mlpackage` in the HF model repo — artifact + the exact script that produced it. Sound good?

beshkenadze:
@lucasnewman hey! Could I ask for a moment of your time to get some clarity on the future of this feature? I've already built an entire pipeline based on ANE & Streaming, so it would be great to know what's in store for it and whether I should continue making changes to the library.

lucasnewman:
@beshkenadze I think this is better as a fork or in your own project rather than part of `mlx-audio`. The inference engines and artifacts are fundamentally different between the libraries, not to mention CoreML will be effectively deprecated starting in macOS 27 in favor of CoreAI. I'm going to close this for now, but for the record I think it's a cool project and you should continue to pursue it on your own terms!

PR #203, opened 2026-06-10
feat(nemotron): cache-aware streaming CoreML/ANE encoder

beshkenadze:
## Cache-aware streaming CoreML/ANE encoder for Nemotron

Runs the Nemotron 3.5 **streaming** FastConformer encoder on the ANE in `generateStream`
(validated uniform-121 feeding + manual cache threading; the prompt MLP and RNN-T decode stay
in MLX). **8-bit palettized** (~28% faster ANE, transcript word-identical). `--stream --ane`
(auto-download) / `--coreml-stream-encoder`.

This is a **power/thermal** feature, not a speed one — and honestly a *partial* GPU offload:
only the encoder moves to the ANE; the decoder stays on the GPU each chunk. For realtime
streaming the speed-vs-MLX difference is moot (RTF ~45–58×); the value is freeing the GPU for
concurrent work + lower power / cooler for always-on / battery streaming.

The incremental streaming session only runs the MLX encoder, so `--ane --stream` routes through
`generateStream` (where the ANE encoder lives); plain `--stream` keeps the low-latency session.
The fixed-shape ANE model only supports the native chunk, so it gains nothing from the session's
variable-chunk latency ladder anyway.

### Known limitation
On very short clips (≲ 2 s, where the whole utterance sits within one native chunk) the streaming
ANE encoder can drop the leading token (e.g. `Intention` → `Tention`) — a chunk-boundary artifact
of the cache-aware streaming path, not the offline ANE encoder (which transcribes such clips
correctly). It does not appear on normal-length speech. Tracked as follow-up.

> ⚠️ **Stacked on #199 + the offline Nemotron PR.** Diff includes those until they merge.
> Kept as **draft** until then.

### DIFF
--- a/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
+++ b/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
@@ -19,6 +19,13 @@ public final class NemotronASRModel: Module, STTGenerationModel {
 
     public var computeDType: DType = .bfloat16
 
+    #if canImport(CoreML)
+    /// Optional fixed-shape CoreML/ANE Conformer encoder for the **offline** path. When set,
+    /// `decode()` runs it instead of the MLX encoder (falls back to MLX on any failure). The
+    /// encoder is a generic fixed-shape Conformer encoder shared with Parakeet (same I/O).
+    public var coreMLEncoder: ConformerCoreMLEncoder?
+    #endif
+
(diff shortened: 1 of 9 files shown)


lucasnewman:
See comment in https://github.com/Blaizzy/mlx-audio-swift/pull/199 -- for now we're not going to include CoreML inference in `mlx-audio`.

PR #204, opened 2026-06-10
Kokoro: fix French nasal vowels dropped by Character tokenization

### LINKED ISSUE #187: [Kokoro] French nasal vowels silently dropped due to Character vs unicodeScalars tokenization
## Summary

Kokoro's tokenizer drops every French nasal vowel (`ɔ̃`, `ɛ̃`, `ɑ̃`, `œ̃`), making French output sound like syllables are missing — e.g. *"bonjour"* synthesizes as *"bjour"*, *"vingt"* as *"vt"*, *"chanson"* as *"chsson"*.

## Reproduction

1. Load `mlx-community/Kokoro-82M-bf16`
2. Generate with `voice: "ff_siwis"`, `language: "fr"`, text `"Bonjour"`
3. Listen — the nasal `[ɔ̃]` is missing

## Root cause

`Sources/MLXAudioTTS/Models/StyleTTS2/Kokoro/KokoroModel.swift`:

```swift
func tokenize(_ text: String) -> [Int] {
    text.compactMap { config.vocab[String($0)] }
}
```

Swift's `for ch in string` (and `compactMap` on a `String`) iterates **`Character`** (extended grapheme clusters), not Unicode scalars.

The IPA lexicon at `beshkenadze/kokoro-ipa-lexicons` emits French nasal vowels as the decomposed pair *base vowel + U+0303 COMBINING TILDE*, e.g. `bonjour\tb ɔ̃ ʒ ˈu ʁ` where `ɔ̃` is `U+0254` + `U+0303`.

Swift fuses that pair into a single `Character` `\"ɔ̃\"`. The Kokoro vocab in `config.json` contains:
- `ɔ` → 76 ✓
- combining tilde → 17 ✓
- precomposed `ɔ̃` → **missing**

`vocab[\"ɔ̃\"]` returns nil → `compactMap` drops the whole grapheme → the nasal vowel disappears from the phoneme token stream → the model never produces the sound.

All four French nasals are affected: `ɔ̃` (bon, son), `ɛ̃` (vin, fin), `ɑ̃` (vent, dans), `œ̃` (un, brun).

## Proposed fix (one line)

```swift
func tokenize(_ text: String) -> [Int] {
    text.unicodeScalars.compactMap { config.vocab[String($0)] }
}
```

Iterate code points so the base vowel and combining mark are looked up independently, producing tokens `[76, 17]` for `ɔ̃` — both of which exist in the vocab.

## Languages affected

French is the obvious one. Same mechanism likely affects any language whose IPA uses combining diacritics: Portuguese (`ã`, `õ`), Polish nasals, anything with tone marks. Worth a regression check against Spanish / Italian / Portuguese — most are unaffected because their lexicon entries don't combine via U+0300-range marks, but worth verifying.

## Verification on demo / Python

Python Kokoro implementations use `espeak-ng` for G2P and operate on byte/code-point streams, never on Swift's grapheme cluster model — that's why the bug only manifests in this Swift port despite the same model weights and same lexicon.

Happy to send a PR with the fix + a small test that asserts `tokenize(\"bɔ̃ʒˈuʁ\").count == 7` (vs 6 today).

slikmedia:
## Summary

Fixes #187.

Kokoro's `tokenize` iterated `String` as `Character` (extended grapheme clusters). French IPA nasal vowels from the lexicon are emitted as base vowel + U+0303 COMBINING TILDE (e.g. `ɔ` + `̃`). Swift fuses that into a single `Character` that is not in `config.vocab`, so `compactMap` silently drops the entire nasal — e.g. "bonjour" phonemizes as `b ɔ̃ ʒ u ʁ` but tokenizes as if `ɔ̃` were missing.

Iterate `unicodeScalars` instead so the base vowel and combining mark are looked up independently; both exist in the vocab.

## Test plan

- [ ] Generate French TTS with `voice: "ff_siwis"`, `language: "fr"`, text `"Bonjour"` — nasal `[ɔ̃]` should be audible
- [ ] Spot-check other languages (EN, ES, IT) for regressions
- [ ] `swift test --filter KokoroTTS`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Sources/MLXAudioTTS/Models/StyleTTS2/Kokoro/KokoroModel.swift
+++ b/Sources/MLXAudioTTS/Models/StyleTTS2/Kokoro/KokoroModel.swift
@@ -114,7 +114,16 @@ public final class KokoroModel: Module, SpeechGenerationModel, @unchecked Sendab
     // MARK: - Tokenizer
 
     func tokenize(_ text: String) -> [Int] {
-        text.compactMap { config.vocab[String($0)] }
+        // Iterate Unicode scalars, not Characters. Swift's `for ch in String`
+        // returns extended grapheme clusters, which fuse combining marks into
+        // their base character. The IPA lexicon emits French nasal vowels as
+        // base vowel + U+0303 COMBINING TILDE (e.g. "ɔ" + "̃"), which Swift
+        // collapses into a single Character "ɔ̃" that isn't in `config.vocab`
+        // — so `compactMap` silently drops the entire nasal vowel and
+        // "bonjour" comes out as "bjour". Iterating scalars preserves the
+        // base + combining pair as two separate tokens, both of which exist
+        // in the vocab. See https://github.com/Blaizzy/mlx-audio-swift/issues/187
+        text.unicodeScalars.compactMap { config.vocab[String($0)] }
     }
 
     // MARK: - SpeechGenerationModel


lucasnewman (approved):
Thanks!

PR #205, opened 2026-06-11
feat(voxtral): true incremental streaming for Voxtral Realtime

beshkenadze:
## Summary

`VoxtralRealtimeModel` ran the **"Realtime"** model **offline**: `generateStream` consumed the whole audio buffer up front (`encodeAndPrefill`) and only then yielded the finished transcript token-by-token. This adds a genuine **online** path — feed audio as it arrives, emit tokens at the model's native transcription delay, O(1) work per chunk.

## What's added

- **`VoxtralRealtimeStreamSession`** (`model.makeStreamSession()` → `step(samples)` / `finish()`): ingests audio incrementally, feeds only newly-frozen conv frames through the transformer encoder with a **persistent per-layer KV-cache**, maintains the decoder KV-cache across calls, emits tokens at the native delay.
- **Incremental encoder** (`encodeIncremental` + sliding-window-aligned block feeding): the key trick is that **RoPE is relative-position invariant**, so feeding conv frames in `sliding_window`-aligned blocks with a **cache reset at each boundary** is bit-exact to both `encodeFull` (≤sw) and `encodeChunked` (>sw) in a single path. Adds a shared `convStemForAudio` / `encodeAudio` seam used by offline + streaming.
- **CLI**: `mlx-audio-swift-stt --stream` routes Voxtral through the session (feeds 480 ms chunks, prints deltas live).
- **Fixture parity test** (compile-guarded — `swift test` can't load the MLX metallib in-sandbox, so MLX runtime verification is done via an executable harness).

## Correctness (WER 0 vs offline)

- The decode loop is already per-audio-token (`inputEmbed = adapterOut[pos] + tokenEmbed`); only `adapterOut` (encoder output) was offline-in-one-shot.
- Conv stem is causal and `prepareMel` right-pads with zeros ⇒ prefix conv frames are frozen (only the trailing partial token is guarded).
- `finish()` reproduces the offline tail zero-pad ⇒ final transcript equals `generate()`.

## Validation (4-bit model, memory-capped 18 GB, M-series)

| Clip | WER vs offline | RTF | per-chunk | realtime |
|---|---|---|---|---|
| intention (1.5 s) | **0.0000** | 1.26 | max 0.452 s | ✅ (cold-start bound) |
| conversational_a (13.26 s) | **0.0000** | **0.649** | median 0.285 s, max 0.455 s (budget 0.480) | ✅ end-to-end + steady-state |

Transcript is bit-identical to `generate()` on both clips.

## Files
- `Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeStreamSession.swift` (new)
- `Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeEncoder.swift` (+`encodeIncremental`)
- `Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift` (`encodeAudio`/`convStemForAudio` seams)
- `Sources/Tools/mlx-audio-swift-stt/App.swift` (Voxtral `--stream`)
- `Tests/MLXAudioSTTTests.swift` (fixture parity test)

## Test plan
- `swift build` ✅ · `swift build --build-tests` ✅
- `mlx-audio-swift-stt --model mlx-community/Voxtral-Mini-4B-Realtime-2602-4bit --audio <clip> --stream` → streams transcript, realtime.

## Follow-ups (not in this PR)
- Incremental conv-stem (raw-sample tail + causal conv caches) to cap per-chunk cost on *minutes*-long streams.
- Expose `--stream-chunk-ms` as a CLI flag (currently 480 ms).

### DIFF
--- a/Sources/Tools/mlx-audio-swift-stt/App.swift
+++ b/Sources/Tools/mlx-audio-swift-stt/App.swift
@@ -363,6 +363,13 @@ enum App {
         audio: MLXArray,
         parameters: STTGenerateParameters
     ) async throws -> STTOutput {
+        // Voxtral Realtime has a true online streaming session — feed audio as it
+        // arrives (480 ms chunks ~ the model's native transcription delay) instead of
+        // the whole-buffer `generateStream`.
+        if let voxtral = model as? VoxtralRealtimeModel {
+            return runVoxtralStreaming(model: voxtral, audio: audio, parameters: parameters)
+        }
+
         var finalOutput: STTOutput?
         var streamedText = ""
         var emittedToken = false
@@ -390,6 +397,24 @@ enum App {
         }
 
         return STTOutput(text: streamedText.trimmingCharacters(in: .whitespacesAndNewlines))
+    }
+
+    /// Drive Voxtral's streaming transcription from a file, printing each delta live.
+    /// The chunked-feed logic lives in `VoxtralRealtimeModel.transcribeStreaming` so other
+    /// callers can reuse it; the CLI only renders the deltas.
+    private static func runVoxtralStreaming(
+        model: VoxtralRealtimeModel,
+        audio: MLXArray,
+        parameters: STTGenerateParameters
+    ) -> STTOutput {
+        var emitted = false
+        let output = model.transcribeStreaming(audio: audio, generationParameters: parameters) { delta in
+            emitted = true
+            print(delta, terminator: "")
+            fflush(stdout)
+        }
+        if emitted { print() }
+        return output
     }
 
     private static func loadModel(repo: String) async throws -> LoadedModel {
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
@@ -210,7 +210,7 @@ public final class VoxtralRealtimeModel: Module, STTGenerationModel {
     }
 }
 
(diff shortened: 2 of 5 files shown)


Sources/Tools/mlx-audio-swift-stt/App.swift
```diff
@@ -363,6 +363,13 @@ enum App {
         audio: MLXArray,
         parameters: STTGenerateParameters
     ) async throws -> STTOutput {
+        // Voxtral Realtime has a true online streaming session — feed audio as it
+        // arrives (480 ms chunks ~ the model's native transcription delay) instead of
+        // the whole-buffer `generateStream`.
+        if let voxtral = model as? VoxtralRealtimeModel {
+            return runVoxtralStreaming(model: voxtral, audio: audio, parameters: parameters)
```
lucasnewman, line 370: Could this be part of the model's implementation? I think calling a different entrypoint is fine, but it doesn't seem right that it's be implemented in the CLI app given other users will want to do the same.

lucasnewman:
@beshkenadze Overall this looks fine, but see my comment about avoiding the extra logic in the CLI app.

beshkenadze:
Moved the logic into the model — `VoxtralRealtimeModel.transcribeStreaming(audio:generationParameters:onDelta:)`. The CLI now just passes a print callback, so the chunked-feed streaming is reusable by any caller.

lucasnewman (approved):
LGTM!

PR #207, opened 2026-06-12
Fix remote tokenizer loading for the MOSS models

### LINKED ISSUE #182: MLXAudioTTS/MossTTSModel.swift error "homeDirectoryForCurrentUser is unavailable for iOS"
Hello: I am found this error "homeDirectoryForCurrentUser is unavailable for iOS" on function below. Not sure how to fix? Anyway this work fine with macOS.

`private static func findCachedHubSnapshot(repo: String) -> URL? {
        let components = repo.split(separator: "/", maxSplits: 1).map(String.init)
        guard components.count == 2 else { return nil }
        let home = FileManager.default.homeDirectoryForCurrentUser
        let snapshotsDir = home
            .appendingPathComponent(".cache/huggingface/hub", isDirectory: true)
            .appendingPathComponent("models--\(components[0])--\(components[1])", isDirectory: true)
            .appendingPathComponent("snapshots", isDirectory: true)
        guard let snapshots = try? FileManager.default.contentsOfDirectory(
            at: snapshotsDir,
            includingPropertiesForKeys: nil
        ) else {
            return nil
        }
        return snapshots.sorted { $0.lastPathComponent > $1.lastPathComponent }.first {
            FileManager.default.fileExists(atPath: $0.appendingPathComponent("config.json").path)
        }
    }`

lucasnewman:
Resolves https://github.com/Blaizzy/mlx-audio-swift/issues/182

### DIFF
--- a/Sources/MLXAudioTTS/Models/MossTTS/MossTTSModel.swift
+++ b/Sources/MLXAudioTTS/Models/MossTTS/MossTTSModel.swift
@@ -78,6 +78,8 @@ public final class MossTTSModel: Module, SpeechGenerationModel, @unchecked Senda
 
     public var tokenizer: MossTTSTextTokenizing?
     public var audioTokenizer: MossAudioTokenizing?
+    private var hfToken: String?
+    private var cache: HubCache = .default
 
     public var sampleRate: Int { config.samplingRate }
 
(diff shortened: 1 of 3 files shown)


Blaizzy (approved):
🚀

PR #208, opened 2026-06-14
feat(nemotron_asr): incremental NemotronASRStreamSession for live mic

beshkenadze:
## What

Adds `NemotronASRStreamSession` — a true **online** streaming session for Nemotron 3.5 ASR. `generateStream` computes the whole-utterance mel up front, so a live caller only gets text once the entire buffer is in. This session ingests audio incrementally:

- `step(_ samples: [Float]) -> Delta` — feed a chunk, get the newly decoded text
- `finish() -> Delta` — flush the tail

Also adds `NemotronASRModel.transcribeStreaming(audio:generationParameters:onDelta:)`, which drives the session over a whole buffer (chunked feed) so any caller can reuse it — the CLI's `--stream` path just renders the deltas. (Mirrors the Voxtral streaming entrypoint from #205.)

## Changes

- **`NemotronASRStreamSession.swift`** (new) — the online session + `transcribeStreaming`
- **`NemotronASRModel.swift`**, **`NemotronASRStreaming.swift`** — extract reusable streaming primitives; no change to the existing offline / `generateStream` behaviour
- **`Tools/mlx-audio-swift-stt/App.swift`** — route `--stream` for Nemotron through `transcribeStreaming`
- **`Tests/MLXAudioSTTTests.swift`** — coverage

## Validation

- `swift build` against current `main` — green.
- Functional: streaming a ~1 min English clip via `step`/`finish` yields a transcript consistent with the offline path.

### DIFF
--- a/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
+++ b/Sources/MLXAudioSTT/Models/NemotronASR/NemotronASRModel.swift
@@ -141,61 +141,18 @@ public final class NemotronASRModel: Module, STTGenerationModel {
             let frameSeconds = Double(self.encoderConfig.subsamplingFactor * self.preprocessConfig.hopLength)
                 / Double(sampleRate)
 
-            var results: [NemoAlignedToken] = []
-            var lastToken = self.blankTokenID
-            var decoderState: NemoLSTMState?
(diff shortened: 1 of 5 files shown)


lucasnewman (approved):
LGTM

PR #209, opened 2026-06-16
Add OmniVoice TTS (NAR diffusion over Qwen3 + HiggsAudioV2 codec, with voice cloning)

JacobLinCool:
Adds [OmniVoice](https://huggingface.co/mlx-community/OmniVoice), a multilingual zero-shot TTS model: a bidirectional NAR diffusion LM over a Qwen3 backbone generating 9 RVQ codebooks at 24 kHz, decoded by a HiggsAudioV2 acoustic codec.

### What's included
- **`Sources/MLXAudioTTS/Models/OmniVoice/`** — model, config, generation params, and the HiggsAudioV2 semantic encode path (HuBERT + SemanticEncoder + fusion + residual RVQ).
- **Three modes:** auto voice, voice design (instruction), and **voice cloning** (`refAudio` + `refText`).
- Registered in `TTSModel.swift`; `Tests/OmniVoiceTests.swift` and a model README included.

### Supporting changes
- **Qwen3 (TTS) backbone rebased on the mlx-swift-lm implementation** — this touches the shared `Models/Qwen3` backbone (also used by VyvoTTS); the public API (`fromPretrained`, `generate`) is preserved. Flagging since it's shared code.
- `ModelUtils`: validate model snapshots recursively.

### Notes for reviewers
- Voice cloning requires the full `mlx-community/OmniVoice` checkpoint; the `-bf16` variant strips `semantic_model.*` and the loader rejects encode with a clear error.
- Encode parity vs the Python reference: 199/200 RVQ frames identical (root flips are fp near-ties); cloned-audio timbre similarity (MFCC) to the Python clone of the same prompt: 0.994.

### DIFF
--- a/Sources/MLXAudioCore/ModelUtils.swift
+++ b/Sources/MLXAudioCore/ModelUtils.swift
@@ -79,19 +79,16 @@ public enum ModelUtils {
             .appendingPathComponent("mlx-audio")
             .appendingPathComponent(modelSubdir)
 
-        // Check if model already exists with required files
+        // Check if model already exists with required files. Some repos keep
+        // their weights in version subfolders (e.g. DeepFilterNet v1/v2/v3),
+        // so search the snapshot recursively rather than only the top level.
         if FileManager.default.fileExists(atPath: modelDir.path) {
-            let files = try? FileManager.default.contentsOfDirectory(at: modelDir, includingPropertiesForKeys: [.fileSizeKey])
-            let hasRequiredFile = files?.contains { file in
-                guard file.pathExtension == normalizedRequiredExtension else { return false }
-                let size = (try? file.resourceValues(forKeys: [.fileSizeKey]))?.fileSize ?? 0
-                return size > 0
-            } ?? false
+            let hasRequiredFile =
+                firstNonEmptyFile(withExtension: normalizedRequiredExtension, under: modelDir) != nil
 
             if hasRequiredFile {
                 // Validate that config.json is valid JSON
-                let configPath = modelDir.appendingPathComponent("config.json")
-                if FileManager.default.fileExists(atPath: configPath.path) {
+                if let configPath = firstNonEmptyFile(withExtension: "json", named: "config.json", under: modelDir) {
                     if let configData = try? Data(contentsOf: configPath),
                        let _ = try? JSONSerialization.jsonObject(with: configData) {
                         print("Using cached model at: \(modelDir.path)")
@@ -132,14 +129,9 @@ public enum ModelUtils {
         )
 
         // Post-download validation: ensure required files are non-zero
-        let downloadedFiles = try? FileManager.default.contentsOfDirectory(
-            at: modelDir, includingPropertiesForKeys: [.fileSizeKey]
-        )
-        let hasValidFile = downloadedFiles?.contains { file in
-            guard file.pathExtension == normalizedRequiredExtension else { return false }
-            let size = (try? file.resourceValues(forKeys: [.fileSizeKey]))?.fileSize ?? 0
-            return size > 0
-        } ?? false
+        // (recursive, for repos that keep weights in subfolders).
+        let hasValidFile =
+            firstNonEmptyFile(withExtension: normalizedRequiredExtension, under: modelDir) != nil
 
         if !hasValidFile {
             Self.clearCaches(modelDir: modelDir, repoID: repoID, hubCache: cache)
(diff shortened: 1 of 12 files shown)


Sources/MLXAudioTTS/Models/Qwen3/Qwen3.swift
```diff
@@ -318,78 +264,57 @@ private class Qwen3ModelInner: Module {
...
+        self.kvHeads = (0..<args.hiddenLayers).map { _ in args.kvHeads }
         self.model = Qwen3ModelInner(args)
 
         if !args.tieWordEmbeddings {
-            self._lmHead.wrappedValue = Linear(args.hiddenSize, args.vocabularySize, bias: false)
+            _lmHead.wrappedValue = Linear(args.hiddenSize, args.vocabularySize, bias: false)
         }
     }
 
-    /// Parse a single row of tokens on CPU
```
lucasnewman, line 351: Why did all the comments in this file get deleted?

Sources/MLXAudioTTS/Models/OmniVoice/README.md
```diff
@@ -0,0 +1,58 @@
+# OmniVoice
+
+Swift/MLX port of [k2-fsa/OmniVoice](https://huggingface.co/k2-fsa/OmniVoice), a multilingual zero-shot TTS model: a bidirectional diffusion LM over a Qwen3 backbone (28 layers, 1024 hidden) generating 9 RVQ codebooks at 24 kHz, decoded by a HiggsAudioV2 acoustic codec.
+
+## Weights
+
+Use the full conversion `mlx-community/OmniVoice` (fp32, includes the complete
+audio tokenizer). `mlx-community/OmniVoice-bf16` ships a stripped tokenizer
+(no `semantic_model.*` weights) and a different fused embedding layout; both
```
lucasnewman: Why does the quantization choice affect whether the tokenizer is stripped down or not? Shouldn't it just always use the full tokenizer?

Sources/MLXAudioCore/ModelUtils.swift
```diff
@@ -79,19 +79,16 @@ public enum ModelUtils {
             .appendingPathComponent("mlx-audio")
             .appendingPathComponent(modelSubdir)
 
-        // Check if model already exists with required files
+        // Check if model already exists with required files. Some repos keep
+        // their weights in version subfolders (e.g. DeepFilterNet v1/v2/v3),
+        // so search the snapshot recursively rather than only the top level.
```
lucasnewman: This change isn't necessarily wrong (although it seems like it might hit some false positives), but I'm curious why this is necessary given both the repos supported have top-level config.json and model.safetensors?

lucasnewman:
@JacobLinCool Thanks! I put a few comments in -- if we can resolve those we can merge.

Sources/MLXAudioTTS/Models/OmniVoice/README.md
JacobLinCool: It's not a choice this PR makes — it's a property of the uploaded checkpoints. The `-bf16` repo's audio tokenizer was exported without the semantic encode path, so the `semantic_model.*` weights simply aren't in the file.

You can see the difference in the file metadata:
- full: https://huggingface.co/mlx-community/OmniVoice/tree/main/audio_tokenizer?show_file_info=audio_tokenizer/model.safetensors — includes `semantic_model.*`
- bf16: https://huggingface.co/mlx-community/OmniVoice-bf16/tree/main/audio_tokenizer?show_file_info=audio_tokenizer/model.safetensors — no `semantic_model.*`

That's an repo uploader decision on the bf16 repo.

Sources/MLXAudioTTS/Models/Qwen3/Qwen3.swift
JacobLinCool: Side effect of rebasing the backbone onto the `mlx-swift-lm` implementation. Restored.

Sources/MLXAudioCore/ModelUtils.swift
JacobLinCool: Accidentally brought in while testing, not needed here. Reverted.

Sources/MLXAudioTTS/Models/OmniVoice/README.md
lucasnewman: @beshkenadze Looks like you uploaded the weights -- any idea why this is the case?

lucasnewman (approved):
🚀

Sources/MLXAudioTTS/Models/OmniVoice/README.md
beshkenadze: @lucasnewman Confirmed and fixed. I've re-converted from k2-fsa/OmniVoice with the current converter (which keeps the semantic weights) and re-uploaded —  OmniVoice-bf16 now has the complete tokenizer (527 tensors, 237 semantic_model.*) and should works for cloning too.

PR #216, opened 2026-07-02
fix(chatterbox): Regular-model reference parity (rounds 1-3)

freeman412:
Brings regular Chatterbox to **blind-ABX parity with the PyTorch reference** (ResembleAI). Confirmed by ear across multiple voices ("I can't tell a difference between any of them").

**Round 1** — CFG uncond pos-emb ordering, `[SPACE]` EnTokenizer tokenization, restore the 56 `attn1.to_out.0.bias` vectors the fp16 checkpoint conversion dropped from the S3Gen flow estimator, CAMPPlus parity (TransitLayer bias, phantom BN).

**Round 2** — `model.train(false)` (BatchNorm was using per-batch stats, collapsing the CAMPPlus x-vector; Dropout was active); S3TokenizerV2 rotary freq exponents `arange(0,dim,2)/dim`; ODE steps 32→10; s3gen prompt-mel double center-pad removed (`stft(center:false)`); HiFT final `leaky_relu` slope 0.1→0.01 (removes the hiss); stop clobbering the global RNG at CFM init (task-local `RandomState`).

**Round 3 (the dominant residual)** — the 16 kHz/24 kHz resampler. The Kaiser resampler cut at Nyquist (rolloff 1.0) while `torchaudio.transforms.Resample` uses rolloff 0.99 + a Hann window, so our resample of the reference clip carried ~**+14 dB** of excess 7–8 kHz energy. CAMPPlus (very sensitive) turned that into a **cos-0.64** speaker x-vector vs the reference, which conditions the flow decoder and brightened every output. Fix: a `sincResampleAudio` that matches torchaudio's windowed-sinc exactly (Hann, width 6, rolloff 0.99) → x-vector **cos 1.0**. Also: regular flow CFM now draws fresh noise per call (was a frozen seed-0 buffer).

Full stage-isolation investigation and per-fix evidence in the write-up. Adds env-gated debug hooks used for the bisection.

PR #217, opened 2026-07-03
fix(chatterbox): Regular-model reference parity (rounds 1-3)

freeman412:
Brings regular Chatterbox to **blind-ABX parity with the PyTorch reference** (ResembleAI) — confirmed by ear across multiple voices. Builds clean (`swift build --product MLXAudioTTS`).

Supersedes #216 (that PR's head branch was recreated after an internal fork merge and decoupled from the PR, so it's stuck at the round-1 commit; this PR carries the complete, build-tested set).

**Round 1** — CFG uncond pos-emb ordering, `[SPACE]` EnTokenizer tokenization, restore the 56 `attn1.to_out.0.bias` vectors the fp16-checkpoint conversion dropped from the S3Gen flow estimator, CAMPPlus parity.

**Rounds 2–3**
- **16k/24k resampler (dominant fix):** replace the Kaiser resampler (cuts at Nyquist, rolloff 1.0) with `sincResampleAudio`, matching `torchaudio.transforms.Resample` exactly (Hann window, `lowpass_filter_width=6`, `rolloff=0.99`). The Kaiser version left ~+14 dB of excess 7–8 kHz energy that corrupted the CAMPPlus speaker x-vector (cos 0.64 → 1.0 vs the reference), which conditions the S3Gen flow decoder and brightened every output.
- **Inference mode:** `model.train(false)` — BatchNorm was using per-batch stats (CAMPPlus' final BN collapsed the x-vector to exact zero) and Dropout was active at generation.
- **Fresh flow noise:** regular CFM now draws `torch.randn_like`-style fresh noise per call; removes the frozen seed-0 `rand_noise` buffer (and with it the `MLXRandom.seed(0)` that clobbered the process-global RNG, making T3 sampling deterministic per run).
- **HiFT:** final `leaky_relu` slope 0.1 → 0.01 (PyTorch's default there) — removes a broadband HF hiss.
- **s3gen prompt mel:** `stft(center: false)` so it isn't double-padded (it already reflect-pads) — was a 2-frame lead vs the reference.

Rotary exponents and the 10-step ODE already match upstream, so they aren't re-touched.


Sources/MLXAudioTTS/Models/Chatterbox/ChatterboxModel.swift
```diff
@@ -947,7 +954,27 @@ public final class ChatterboxModel: Module, SpeechGenerationModel, @unchecked Se
...
-        let sanitizedWeights = model.sanitize(weights: weights)
+        var sanitizedWeights = model.sanitize(weights: weights)
+
+        // The mlx-community Regular conversion dropped the flow estimator's 56
+        // trained `attn1.to_out.0.bias` vectors (present in the original ResembleAI
+        // torch checkpoint and in the Turbo conversion). Restore them from the
+        // bundled sidecar; keys are in Swift module naming so they merge directly.
+        // Only fills keys the checkpoint doesn't provide, so a future corrected
+        // checkpoint wins automatically.
+        if let sidecarURL = Bundle.module.url(
```
lucasnewman, line 965: We need to re-convert the model, we aren't going to patch over it with local artifacts. `mlx-community` is self-add on HuggingFace, so you can just upload a corrected version of the weights if needed.

PR #219, opened 2026-07-05
OmniVoice: emit per-step denoise progress through generateStream

raghakot:
## Summary

OmniVoice generates NAR-diffusion audio in one shot, so `generateStream` consumers currently get **no signal until the final `.audio` event** — for a 30s utterance that's ~30s of silence with no way to render progress. Unlike AR models, though, OmniVoice's denoise loop has a **deterministic step count**, so exact (not estimated) progress is available for free.

This PR surfaces it:

- `AudioGeneration` gains a `case progress(Double)` (fraction 0–1), documented as exact-step progress for models with deterministic step counts.
- `OmniVoiceModel.generateAudio` takes an optional `onStepProgress: (@Sendable (Int, Int) -> Void)?` and invokes it after each denoise step (post-`eval`, so the step's compute is actually done when the event fires).
- The streaming `generateStream(…, streamingInterval:)` overload yields `.progress(done/total)` through the existing continuation.

## Compatibility

- Additive enum case: `proxyAudioStream` / `generateSamplesStream` filter with `guard case .audio` and are unaffected; the one exhaustive switch in-tree (`Sources/Tools/mlx-audio-swift-tts/App.swift`) is updated.
- `generateAudio`'s new parameter is defaulted; the non-streaming `generate` paths are unchanged.
- No behavior change for any other model family.

## Testing

- `swift build --target mlx-audio-swift-tts` clean.
- Ran the package CLI (`--benchmark` streaming path) against `mlx-community/OmniVoice-bf16` with a temporary print on `.progress`: exactly 32 events per generation, monotonic `0.03125 → 1.0`, including generations where trailing schedule steps have an empty unmask quota (those emit too, so the fraction always reaches 1.0). Audio output unchanged.

Happy to also thread progress through other diffusion-style models (Irodori's flow matching, etc.) in a follow-up if this shape looks right.

### DIFF
--- a/Examples/VoicesApp/VoicesApp/ViewModels/TTSViewModel.swift
+++ b/Examples/VoicesApp/VoicesApp/ViewModels/TTSViewModel.swift
@@ -453,6 +453,14 @@ class TTSViewModel {
                                 generationProgress = "Generated \(chunkTokenCount) tokens..."
                             }
                         }
+                    case .progress(let fraction):
+                        // Diffusion models report exact step progress instead of tokens.
+                        let percent = Int((fraction * 100).rounded())
+                        if chunks.count > 1 {
+                            generationProgress = "Chunk \(index + 1)/\(chunks.count): \(percent)%"
+                        } else {
+                            generationProgress = "Generating... \(percent)%"
+                        }
                     case .info(let info):
                         tokensPerSecond = info.tokensPerSecond
                     case .audio(let audioData):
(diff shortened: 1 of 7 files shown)


lucasnewman:
@raghakot Looks like tests are failing.

raghakot:
Sorry, this fell off my radar. Wasn't the progress change itself; `main` picked up a few new exhaustive switches over `AudioGeneration` after I opened this. Merged main and fixed all six. Tests pass locally.

lucasnewman (approved):
Thanks!

PR #225, opened 2026-07-15
Quantized KV cache for MOSS-Transcribe-Diarize (opt-in, memory-bounded long-form)

iannellomarco:
## What this enables

Adds **opt-in quantized KV cache** support to `MOSS-Transcribe-Diarize`. On a small-unified-memory Mac, the full-precision KV cache is what caps how long a clip you can transcribe in a single pass before you're forced to chunk (and lose cross-chunk context). Quantizing the retained context lets the same machine do long single-pass transcription within a much smaller memory budget.

Everything here is **off by default** — with no new parameters set, decoding is bit-for-bit identical to today.

## Design

- **Dispatcher swap (not a reimplementation).** `Qwen3ASRTextAttention` now routes through `MLXLMCommon.attentionWithCacheUpdate` instead of a manual `cache.update` + `scaledDotProductAttention`. This is what makes a `QuantizedKVCache` usable on this path at all — previously it would hit a `fatalError`. RoPE is still applied against the pre-update cache offset, so the standard (non-quantized) path is unchanged. This mirrors how the other Qwen3-based models in the repo already call `attentionWithCacheUpdate`.
- **Protocol-based audio-injection gate.** The prefill audio-feature injection in the MOSS backbone was gated on a concrete `KVCacheSimple` downcast; it now reads `KVCache.offset` directly, so injection fires correctly for any cache type at offset 0 (behavior-preserving for the existing dense path).
- **Correct masking for quantized multi-token steps.** Multi-token forward steps over a quantized cache need an explicit additive float mask — the quantized attention path substitutes a near-zero constant (not `-inf`) for masked positions of a boolean/symbolic mask, which would let future positions leak into the softmax. A small helper builds the additive mask only in that case and otherwise defers to the standard helper.
- **Opt-in parameters, safe defaults.** `STTGenerateParameters` gains `kvBits: Int? = nil`, `kvGroupSize: Int = 64`, `quantizedKVStart: Int = 0`, threaded through `generate` / `generateSingleChunk` / `generateTokenIds`. `kvBits == nil` leaves the cache at model precision. Quantization happens *after* prefill via `MLXLMCommon.maybeQuantizeKVCache`.
- **Bounded prefill memory.** For a long prompt, the full-precision KV would otherwise spike transiently before it's ever converted, so the cache is quantized incrementally inside the chunked prefill loop (matching mlx-lm's chunked-prompt behavior). When quantization is active the prefill chunk size is also reduced (unfused quantized attention materializes a `[heads, chunk, context]` score tensor), which bounds that spike; the full-precision path keeps its original fused kernel and chunk size.
- **Pre-quantized checkpoint loading.** `MossTranscribeDiarizeConfig` decodes a `quantization` / `quantization_config` block (plus per-layer overrides), and `fromModelDirectory` swaps only the layers that ship quantized tensors — the audio tower and adaptor stay dense — mirroring the sibling models in the repo.

## Tooling

- Adds a `mlx-audio-swift-moss-tools` executable under `Sources/Tools/` (same convention as the existing `mlx-audio-swift-*` CLIs): `quantize` converts a local bf16 MOSS checkpoint to MLX-quantized safetensors that `fromModelDirectory` loads directly, and `bench` decodes one 16 kHz WAV in a single pass with optional KV quantization and writes a JSON report. Local directories only, no network.
- Bumps the pinned `mlx-swift` / `mlx-swift-lm` to `0.31.4` / `3.31.4` (already within the existing `upToNextMajor` constraints) — this is the KV-cache revision the quantized path was validated against. Happy to drop this commit if you'd rather bump `Package.resolved` yourself.

## Validation

Validated on a 56-minute multi-speaker meeting recording:

- **8-bit / group size 64 KV** produces a transcript matching the bf16 reference (same speaker structure) at **~30% lower peak memory**.
- **4-bit / group size 32 KV degrades into repetitive output on hour-scale audio and should not be used for long inputs** — it's plumbed through but is not a safe default, hence the honest caveat here.

Unit tests cover config decoding (both `quantization` and `quantization_config` keys, and the dense-by-default case) and the `STTGenerateParameters` defaults.

## Compatibility

**No breaking changes.** All new parameters default to values that reproduce today's exact behavior; existing call sites and the dense checkpoint path are untouched. `swift build` passes and the `MossTranscribeDiarizeModuleSetupTests` suite (including the existing streaming/timestamp tests) is green.

The quantized-attention wiring came out of building a macOS transcription app on top of this library; contributing it back since it's generally useful for long-form MOSS transcription on constrained hardware.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "0de47e2efcfdbd2abb306b485c46d6fc9b94854ac10c7413ede360d5ad7e2260",
+  "originHash" : "4f148e1f0e3958fd3f86560361e4a93ad32a6877d96593d3a361ee25e86da198",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -24,8 +24,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/ml-explore/mlx-swift.git",
       "state" : {
-        "revision" : "61b9e011e09a62b489f6bd647958f1555bdf2896",
-        "version" : "0.31.3"
+        "revision" : "dc43e62d7055353c7f99fa071a4e71d29dfddc44",
+        "version" : "0.31.4"
       }
     },
     {
@@ -33,8 +33,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/ml-explore/mlx-swift-lm.git",
       "state" : {
-        "revision" : "1c05248bb0899e2a7a4962b84d319cf12f4e12aa",
-        "version" : "3.31.3"
+        "revision" : "bd4b7434e6bdb588c7ef55706ff8904cb7fd4c57",
+        "version" : "3.31.4"
       }
     },
     {
@@ -240,8 +240,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swiftlang/swift-syntax.git",
       "state" : {
-        "revision" : "0687f71944021d616d34d922343dcef086855920",
-        "version" : "600.0.1"
+        "revision" : "79e4b74a295b6eb74a8b585e3a39d29e70c1dbd1",
+        "version" : "603.0.2"
       }
     },
     {
--- a/Sources/MLXAudioSTT/Generation.swift
+++ b/Sources/MLXAudioSTT/Generation.swift
@@ -11,8 +11,7 @@ public struct STTGenerateParameters: Sendable {
     public let minChunkDuration: Float
     public let repetitionPenalty: Float
     public let repetitionContextSize: Int
-    /// KV-cache quantization bits. `nil` keeps the cache at model precision
-    /// (bit-for-bit identical to prior behavior).
+    /// KV-cache quantization bits; `nil` keeps model precision.
     public let kvBits: Int?
     /// Group size for KV-cache quantization.
     public let kvGroupSize: Int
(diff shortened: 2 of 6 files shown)


lucasnewman:
@iannellomarco Thanks! I think this is reasonable. We don't need the additional cli tools right now and typically handle conversion in the python repo, so please remove the new tool.

The Qwen3-ASR changes look fine but also unrelated, so if you could break those into a separate PR that would be better as well.

Once those changes are in place we can merge. Make sure your commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) so we can clear the merge requirements.

iannellomarco:
Thanks for the review! All three items addressed:

- **CLI removed** — the `moss-tools` target and its commit are gone from this branch; conversion stays in the Python repo as you prefer.
- **Qwen3-ASR change split out** → #228.
- **Commits signed** — the branch was rebuilt onto current `main` with signed commits.

One ordering note: this PR's opt-in `kvBits` path reaches the cache through the attention dispatcher that #228 introduces (the MOSS backbone is `Qwen3ASRTextModel`), so #228 should merge first — otherwise quantized decode here would hit the `fatalError` in `QuantizedKVCache.update()`. With quantization off (the default), both are behavior-neutral in any order.

iannellomarco:
With #228 merged (thanks @lucasnewman!), the ordering constraint here is satisfied — the dispatcher this PR's quantized path relies on is now in `main`, so this can land without ever exposing a crashing path. The branch is conflict-free against current `main`; happy to rebase or squash if you'd prefer a tidier history. No rush, and thanks again for the reviews.

lucasnewman (approved):

PR #227, opened 2026-07-15
OmniVoice: honor task cancellation during generation

raghakot:
OmniVoice is currently the only TTS model in the repo whose generation loop never checks for cancellation — Qwen3TTS, EchoTTS, PocketTTS, Qwen3, and the other models all call `try Task.checkCancellation()` inside their sampling/diffusion loops. As a result, cancelling a consumer mid-generation leaves the full diffusion compute running to completion (tens of seconds of GPU work for long utterances) with the result silently dropped.

This PR aligns OmniVoice with the existing convention:

- `try Task.checkCancellation()` at the top of each diffusion step in `generateAudio` (matching `EchoTTSSampling`'s per-step check), and once more before the vocoder decode.
- `generateStream` now holds its compute `Task` and cancels it from `continuation.onTermination`, so dropping the stream actually stops the work (same effect as the proxy-stream wiring in `Generation.swift`).

Purely additive: non-cancelled generations execute identically. Verified on an M4 Mac and an iPhone 17 Pro Max — cancelling mid-generation now interrupts within one diffusion step (~1 s for a 30-word utterance) instead of running the chunk to completion.


lucasnewman (approved):
Thanks!

PR #229, opened 2026-07-17
VoxtralRealtime streaming: stop clearing the Metal buffer pool on every step

T0mSIlver:
`VoxtralRealtimeStreamSession.advance()` calls `Memory.clearCache()` on both exit paths — 10×/s at a live 100 ms feed cadence. Every step re-allocates its working set cold from the Metal driver, and any `Memory.cacheLimit` the host sets is defeated (the pool never lives long enough to be bounded by it). Field-observed as a 10 Hz RSS sawtooth between the ~3.4 GB weight floor and 5+ GB.

The offline paths already handle this right: `generate()` / `generateStream()` clear every 256 tokens. This PR makes the streaming session match — clear every 256 decoded tokens plus once on the `finish()` flush, otherwise leave pool management to the host via `Memory.cacheLimit` (the mlx-lm / voxmlx policy). No computation changes.

## Evidence

M1 Pro, Voxtral-Mini-4B 4-bit, 60 s stream at 100 ms cadence, `cacheLimit` 4 GB; measured together with #230:

| mark | step ms (main) | step ms (this + #230) |
|---|---|---|
| 5s | 183.2 | 80.1 |
| 15s | 162.7 | 74.2 |
| 30s | 165.8 | 76.6 |
| 60s | 162.5 | 80.3 |

`main` streams 60 s of audio in 114.7 s (1.9× slower than realtime); fixed is flat at ≈0.76 RTF with the pool held at the 4 GB limit (memory moves only at the intended 256-token clears). Word accuracy on a live-audio integration suite unchanged (0.789).

Companion PR: #230 (independent, merge cleanly together). Staged with full evidence at T0mSIlver#2.

### DIFF
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeStreamSession.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeStreamSession.swift
@@ -166,8 +166,13 @@ public final class VoxtralRealtimeStreamSession {
         // re-allocated cold from the Metal driver 10x/s in live streaming and defeats
         // any `Memory.cacheLimit` the host process sets. Match the offline `generate`
         // loop instead: clear every 256 decoded tokens (see `decode`), plus once when
-        // the utterance finishes so memory returns to the weight floor between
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
LGTM

PR #230, opened 2026-07-17
VoxtralRealtime streaming: incremental mel/conv front end (O(N²) → O(N) per utterance)

T0mSIlver:
The streaming session is incremental in its transformer but not its front end: every `step()` re-runs mel + both conv layers over the ENTIRE accumulated audio (plus ~3.9 s of constant pad), including a per-sample CPU reflect-pad loop and full-buffer GPU↔CPU copies. That is O(N²) per utterance, and per-step latency grows with utterance length.

This PR makes the front end incremental — mirroring the Python reference (voxmlx), generalized to arbitrary chunk sizes:

- **`VoxtralRealtimeMelStream`** — carries the not-yet-framed sample suffix and emits a mel column only once its full 400-sample window is buffered, so every emitted frame is final and matches offline framing for any feed pattern.
- **`VoxtralRealtimeConvStemState` / `convStemStep`** — conv1 (k3 s1) carries its last 2 input frames; conv2 (k3 s2) carries a 1–2 frame suffix aligned to the next stride-2 window, so the downsample phase survives odd-sized chunks.
- **`advance()`** feeds only new samples onward; no full-buffer recompute, copies, or retained raw audio. `finish()` reproduces the offline right pad exactly; `step()` after `finish()` is ignored (previously it silently corrupted state); empty-audio `finish()` matches offline `generate`. Offline/batch paths untouched.

## Exactness

Offline mel frame t covers samples [160t−200, 160t+200), and both reflect pads reduce to zeros under the model's own padding — so the carried tails re-present exactly the windows the offline code reads, and the streaming freeze boundary never consumes rows that depend on the provisional right pad (full argument in the commit messages). An independent review re-derived the arithmetic and confirmed it numerically across pad configs × audio lengths × chunk patterns (down to 1-sample chunks), including identical per-step token emission timing for every sample count 0–60000.

## Tests

`VoxtralRealtimeStreamingFrontEndTests` — 10 deterministic cases: irregular chunk mixes, single giant chunk, step-after-finish, double-finish, non-zero transcription delay, a sliding-window-crossing run with a real attention layer, empty-audio finish. Token comparisons are exact; the mel/conv comparisons allow a tiny epsilon because MLX kernels are not per-row deterministic across batch shapes (CI's virtualized Metal shows ~2e-4 log-mel drift that hardware Metal does not). The offline side of every comparison is the untouched batch code, so nothing passes vacuously.

## Evidence

M1 Pro, Voxtral-Mini-4B 4-bit, 60 s stream at 100 ms cadence; measured together with #229:

| mark | step ms (main) | step ms (this + #229) |
|---|---|---|
| 5s | 183.2 | 80.1 |
| 15s | 162.7 | 74.2 |
| 30s | 165.8 | 76.6 |
| 60s | 162.5 | 80.3 |

`main` streams 60 s of audio in 114.7 s; with these PRs latency is flat (80 ms @60 s ≈ 80 ms @5 s), ≈0.76 RTF. Word accuracy on a live-audio integration suite unchanged (0.789).

Companion PR: #229 (independent, merge cleanly together). Staged with full evidence at T0mSIlver#3.

### DIFF
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtime.swift
@@ -376,8 +376,8 @@ extension VoxtralRealtimeModel {
     /// Causal conv + causal/sliding-window attention make `adapterOut[0..<k]`
     /// identical whether encoded from a prefix or the full buffer.
     /// Conv-stem seam: audio → conv-stem frames (`convOut`, the transformer input),
-    /// the per-token span, and the prompt length. Shared by the offline encode and
-    /// the streaming session (which feeds `convOut` incrementally).
+    /// the per-token span, and the prompt length. Used by the offline encode; the
+    /// streaming session builds the same rows incrementally (`convStemStep`).
     func convStemForAudio(
         audio: MLXArray,
         transcriptionDelayMs: Int?
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeAudio.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeAudio.swift
@@ -72,6 +72,36 @@ enum VoxtralRealtimeAudio {
         return logSpec
     }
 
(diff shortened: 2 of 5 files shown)


lucasnewman (approved):
Thanks!

PR #231, opened 2026-07-17
Hoist per-layer-invariant attention inputs out of the Voxtral layer loops

T0mSIlver:
Every Voxtral Realtime transformer layer rebuilt identical per-forward values inside its attention: the interleaved-RoPE cos/sin tables (encoder and decoder) and the SDPA sliding-window mask (encoder cached paths) — O(nLayers) copies of the same tiny kernels per pass, pure launch overhead in streaming where a pass covers a few new frames or a single decoded token. The encoder now builds a shared `VoxtralRealtimeEncoderAttentionInputs` once per pass in `encodeFull` / `encodeChunked` / `encodeIncremental`; the decoder computes its RoPE tables once per forward. The offline/batch path gets the same reduction.

Sharing one mask relies on the per-layer KV caches advancing in lockstep, which is how all three encode paths drive them; the builder `precondition`s it (including nil-ness uniformity). This also matches the Python reference (voxmlx), which builds mask and RoPE once per forward — the per-layer rebuild was porting overhead, not intent.

Outputs are unchanged: same operations, same inputs, same order — only the number of identical subgraph instantiations changes, so every layer consumes bit-identical arrays. Covered by the existing streaming/offline equivalence tests; no public API changes.

## Evidence

M1 Pro, 60 s stream at 100 ms cadence, Voxtral-Mini-4B-Realtime 4-bit: encoder graph build 3.5–5.0 → 1.7–2.1 ms/step, whole step ~99.8 → ~96.6 ms (~3%). Modest but strict — fewer kernel launches for identical outputs, simpler per-layer attention bodies, and the lockstep invariant now explicit and loudly enforced.

Notes: textually conflicts with #230 in `VoxtralRealtimeEncoder.swift` (content-independent) — happy to rebase whichever lands second. Staged at T0mSIlver#4.

### DIFF
--- a/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeDecoder.swift
+++ b/Sources/MLXAudioSTT/Models/VoxtralRealtime/VoxtralRealtimeDecoder.swift
@@ -80,7 +80,7 @@ final class VoxtralRealtimeDecoderAttention: Module {
     /// launches per decoded token. The operations are unchanged, so the tables
     /// (and therefore the layer outputs) are bit-identical to the previous
     /// per-layer construction.
-    static func ropeFrequencies(
+    fileprivate static func ropeFrequencies(
         positions: MLXArray,
         headDim: Int,
         ropeTheta: Float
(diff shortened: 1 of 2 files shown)


lucasnewman:
@T0mSIlver Thanks, can you take a look at the merge conflicts with your other changes which are now merged?

T0mSIlver:
@lucasnewman Rebased onto main (542fffa) — the Encoder.swift conflict with #230 is resolved, keeping both `attentionInputs` and `convStemStep`. Shows as mergeable now. Anything else you'd like changed here?

lucasnewman (approved):
Thanks!

PR #233, opened 2026-07-22
Add reproducible OmniVoice generation seed

yanchenko:
## Summary

- add an optional `UInt64` seed to `OmniVoiceGenerateParameters`
- seed MLX before OmniVoice preprocessing and diffusion when callers request reproducible generation
- preserve current random-state behavior when the seed is `nil`
- document the setting and cover its default and custom values

## Testing

- `swift test --filter OmniVoiceTests`
- `xcodebuild build-for-testing -scheme MLXAudio-Package -destination 'platform=macOS' MACOSX_DEPLOYMENT_TARGET=14.0 CODE_SIGNING_ALLOWED=NO`


lucasnewman (approved):
Thanks!

PR #234, opened 2026-07-22
Honor Chatterbox emotion override with default conditioning

yanchenko:
## Summary

- apply `emotionAdvOverride` when Chatterbox synthesizes with bundled default conditioning
- preserve the checkpoint emotion value when no override is set
- add a focused regression test for both paths

Previously the override was only consulted while preparing reference-audio conditioning, so callers using the default voice could set it without affecting synthesis.

## Testing

- `xcodebuild build-for-testing -scheme MLXAudio-Package -destination 'platform=macOS' MACOSX_DEPLOYMENT_TARGET=14.0 CODE_SIGNING_ALLOWED=NO`\n- `xcodebuild test-without-building -scheme MLXAudio-Package -destination 'platform=macOS' -only-testing:MLXAudioTests/ChatterboxGenerationTests -parallel-testing-enabled NO CODE_SIGNING_ALLOWED=NO`

### DIFF
--- a/Sources/MLXAudioTTS/Models/Chatterbox/ChatterboxModel.swift
+++ b/Sources/MLXAudioTTS/Models/Chatterbox/ChatterboxModel.swift
@@ -18,6 +18,10 @@ import MLXNN
 @preconcurrency import MLXLMCommon
 import Tokenizers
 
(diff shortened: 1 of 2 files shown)


lucasnewman:
@yanchenko Good catch, thanks!

lucasnewman (approved):

PR #235, opened 2026-07-26
Fix quantized Whisper checkpoint loading

Parassharmaa:
## Summary

- quantize Whisper linear layers and the tied decoder embedding before loading packed checkpoints
- map tied embedding scales and biases from mlx-whisper checkpoints
- route vocabulary projection through the embedding module so quantized weights dequantize correctly
- add dense and quantized tied-projection tests

## Validation

- `swift build --product mlx-audio-swift-stt`
- transcribed Japanese and English FLEURS clips with `mlx-community/whisper-large-v3-turbo-asr-4bit`; outputs matched the Python MLX checkpoint
- `swift test --filter Whisper` reached test compilation but the selected Command Line Tools environment lacks the Swift `Testing` module


lucasnewman (approved):
LGTM

PR #237, opened 2026-07-29
Stream Fish Speech progressively

steipete:
## Summary

- stream Fish Speech audio as each bounded text segment finishes
- preserve exact plain-text whitespace and speaker markers across batches
- propagate cancellation to the producer without retaining emitted MLX arrays
- keep request-level generation statistics and clarify the existing chunk event contract

## Why

Fish Speech currently ignores `streamingInterval`, generates every segment, concatenates the full waveform, and only then emits one audio event. Consumers therefore cannot begin playback early, and canceling the consumer does not provide a useful progressive boundary.

This keeps the buffered `generate` path unchanged. The streaming path uses byte-bounded, Unicode-aware text batches, yields each decoded segment immediately, and emits one aggregate info event when the request completes.

## Proof

```text
xcodebuild test -scheme MLXAudio-Package \
  -destination 'platform=macOS,arch=arm64' \
  -only-testing:MLXAudioTests/FishSpeechTests

14 tests passed
```

The focused suite covers whitespace preservation, CJK/unspaced input, grapheme boundaries, oversized graphemes, long speaker-tagged turns, invalid intervals, and existing Fish model behavior.

### DIFF
--- a/Sources/MLXAudioCore/Generation/GenerationTypes.swift
+++ b/Sources/MLXAudioCore/Generation/GenerationTypes.swift
@@ -52,7 +52,7 @@ public enum AudioGeneration: Sendable {
     case token(Int)
     /// Generation statistics
     case info(AudioGenerationInfo)
-    /// Final generated audio
+    /// Generated audio chunk. Concatenate repeated values for the complete waveform.
     case audio(MLXArray)
 }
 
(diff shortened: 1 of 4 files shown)


lucasnewman (approved):
Thanks!

PR #239, opened 2026-07-31
Kokoro: expose the per-phoneme durations `generate` already computes

SAPSentinel:
`KokoroModel.callAsFunction` predicts per-phoneme durations on every synthesis and returns them as `predDur`, but `generate` discards them:

```swift
let (audio, _) = self.callAsFunction(inputIds: inputIds, refS: refS, speed: speed)
return audio.reshaped([-1])
```

Callers doing lip-sync, viseme animation or subtitle alignment then have to re-derive timing from the waveform — strictly less accurate than the alignment the model just used to *produce* that waveform.

This adds `generateWithDurations(...)`, returning `(audio, phonemes, durations)`. `generate` now delegates to it and is unchanged for existing callers.

Notes:

- The phonemized string is returned alongside the durations because durations alone can't be mapped back to phonemes — `generate` already computes it internally and drops it too.
- Durations are in acoustic frames. The hop length isn't exposed, but callers can derive seconds exactly: the utterance spans `durations.sum()` frames and `audio.count / sampleRate` seconds. Documented on the method.
- The token array is padded with a leading and trailing `0`, so `durations` is two longer than the phoneme count. Also documented.

Purely additive; no behaviour change. Builds clean against `main` (`swift build --target MLXAudioTTS`).

Context: driving avatar visemes on iOS.

Happy to follow up with a test if you'd like one — I held off because covering this properly needs a model download, and I wasn't sure how you'd prefer that handled in CI.

### DIFF
--- a/Sources/MLXAudioTTS/Models/StyleTTS2/Kokoro/KokoroModel.swift
+++ b/Sources/MLXAudioTTS/Models/StyleTTS2/Kokoro/KokoroModel.swift
@@ -136,6 +136,42 @@ public final class KokoroModel: Module, SpeechGenerationModel, @unchecked Sendab
         language: String?,
         generationParameters: GenerateParameters
     ) async throws -> MLXArray {
+        try await generateWithDurations(
+            text: text, voice: voice, refAudio: refAudio, refText: refText,
+            language: language, generationParameters: generationParameters
+        ).audio
+    }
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
@SAPSentinel Thanks!

PR #247, opened 2026-08-19
Fix Qwen3-ASR mel frontend: Slaney mel scale + periodic Hann window

ZigZagYea:
## Problem

Qwen3ASR's audio frontend (`preprocessAudio` → `MLXAudioCore.computeMelSpectrogram`) produced mel features that don't match the reference implementation, causing noticeable quality degradation on noisy real-world audio:

- `computeMelSpectrogram` built the mel filterbank with the **HTK mel scale** (the `melFilters` default), while transformers' `WhisperFeatureExtractor` (which Qwen3-ASR's `preprocessor_config.json` points to, and which mlx-audio **Python** uses) uses the **Slaney** mel scale.
- `hanningWindow` produced a **symmetric** Hann window, while `torch.hann_window` (used by the reference) is **periodic**.

Clean speech (TTS, quiet recordings) survives the mismatch fine, which makes it easy to miss. On a noisy 108-minute real meeting we saw hard-phrase misrecognitions from all quantization variants, e.g. the spoken phrase `集团会开权限给我们` ("the group will grant us permissions") came out as `经常会开一些黑幕` ("...dark secrets") — reproducibly, across 4bit/6bit/8bit.

## Fix

Add opt-in parameters (defaults preserve current behavior for every other caller):

- `hanningWindow(size:periodic:)` — `periodic: true` matches `torch.hann_window`
- `computeMelSpectrogram(..., melScale:, hannPeriodic:)` — passes through to the filterbank/window

…and pass `melScale: .slaney, hannPeriodic: true` from `Qwen3ASR.preprocessAudio`.

## Verification

On the same audio (noisy real meeting segments), after the fix the Swift output matches **mlx-audio Python** (which uses `transformers.WhisperFeatureExtractor`) **word-for-word**, and the misrecognized phrase above is transcribed correctly. Speed and memory are unchanged (RTF 0.042–0.063, ~1.17GB RSS for 0.6B-8bit on M4 Pro).

| sample | before | after (= Python reference) |
|---|---|---|
| 8s noisy | `……它的保险，就是由银行所担。` | `……它的附件就只有银行回单。` |
| 16s noisy | `……嗯，经常会开一些黑幕。` | `……嗯，集团会开权限给我们。` |

Note: the same HTK-vs-Slaney / symmetric-vs-periodic mismatch likely affects every model that goes through `computeMelSpectrogram` with Whisper-style front ends; I kept this change opt-in to avoid changing behavior for other models without verification.

## Testing

- `swift build` passes; the new fixture assertions were verified locally via a scratch executable against the values generated by `WhisperFeatureExtractor` (CLT-only environments can't run `swift test` without Xcode — the included `Tests/MelFrontendReferenceTests.swift` runs in CI).

### DIFF
--- a/Sources/MLXAudioCore/DSP.swift
+++ b/Sources/MLXAudioCore/DSP.swift
@@ -12,9 +12,12 @@ import MLX
 
 
 /// Create a Hanning window of given size.
-public func hanningWindow(size: Int) -> MLXArray {
+/// `periodic: true` matches `torch.hann_window` (librosa/Whisper default);
+/// the previous symmetric-only form is kept as default for compatibility.
+public func hanningWindow(size: Int, periodic: Bool = false) -> MLXArray {
+    let effectiveSize = periodic ? size + 1 : size
+    let denom = Float(effectiveSize - 1)
     var window = [Float](repeating: 0, count: size)
-    let denom = Float(size - 1)
     for n in 0..<size {
         window[n] = 0.5 * (1 - cos(2 * Float.pi * Float(n) / denom))
     }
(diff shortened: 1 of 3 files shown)


lucasnewman (dismissed):
Thanks!

lucasnewman:
@ZigZagYea Looks like tests are failing, can you take a look?

ZigZagYea:
Thanks for flagging this! The failure was caused by a missing `import MLX` in the new test file. I’ve fixed it in `0307a89` and verified that the full test workflow passes on my fork: https://github.com/ZigZagYea/mlx-audio-swift/actions/runs/33737642193

lucasnewman (approved):

PR #253, opened 2026-08-31
FireRedASR2: fix beam search topK gathering across rows

Kn0688:
The 2D branch of topK used plain `take`, which broadcasts the flattened indices across every row: with beam > 1 all candidate tokens were scored with beam 0's logits. Use `takeAlong` so each index row gathers from the matching input row.

Adds characterization tests pinning topK semantics and the toy-model beam search output (pre-fix output was [3, 3, 3, 3, 3, 3, 4, 4]).


lucasnewman (approved):
Thanks!

PR #256, opened 2026-09-02
STT: scale fbank input to int16 unconditionally in FireRedASR2/SenseVoice

Kn0688:
Both models' Kaldi fbank recipe expects int16-scale audio, and the input was auto-detected with `amplitude <= 1.0`. Lossy decoders (AAC via AVFoundation) can overshoot past 1.0 on clipped content, which flipped the decision: the x32768 scaling was skipped, features landed far off the training distribution, and decoding silently collapsed — on iOS, over half of the VAD segments of a long podcast decoded empty while the same audio (peak 0.999969 via ffmpeg) was fine on Mac.

Scale unconditionally instead, matching sherpa-onnx (which hard-codes normalize_samples = false for FireRedASR) and the existing MossFormer2SE and FSMNVAD code. Input contract is normalized float in [-1, 1], as produced by loadAudioArray.

Adds a regression test with clipped input (max amplitude > 1).

### DIFF
--- a/Sources/MLXAudioSTT/Models/FireRedASR2/FireRedASR2Audio.swift
+++ b/Sources/MLXAudioSTT/Models/FireRedASR2/FireRedASR2Audio.swift
@@ -18,10 +18,15 @@ enum FireRedASR2Audio {
             return MLXArray.zeros([0, numMels], type: Float.self)
         }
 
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
Thanks!

PR #263, opened 2026-09-17
VoxtralRealtime streaming: drop conv and adapter rows once they are consumed

T0mSIlver:
A `VoxtralRealtimeStreamSession` kept every conv-stem row and every adapter row for the life of the stream, and each step concatenated the new rows onto that history. Memory and per-step copy time grew for as long as audio kept coming. On Voxtral Mini 4B Realtime, active memory was still rising about 23 MB per minute after the decoder window had filled.

This is one of three PRs that keep step time and memory flat over hour-long streams, with #264 (decoder KV cache) and #265 (transcript text). None of them changes a decoded token.

## Changes

- The encoder only reads conv rows from `encState.consumed` on, and the decoder only reads adapter rows from `decPos` on. After each step, `trimRetainedTails` drops the rows below those positions. `convRowsDropped` and `adapterRowsDropped` count them, so frame and decode positions stay absolute.
- `feedIncremental` takes `startIndex`, the absolute index of the first row it is given.
- Prefill reads the first adapter rows while `decPos` is still 0, so nothing it needs is dropped. A precondition checks that.
- `step` ignores samples once the stream has finished on EOS or `maxTokens`. Before, they piled up in `pendingSamples` and nothing read them.

## Tests

- `longStreamMatchesOfflineWithBoundedBuffers` streams a minute of audio in irregular chunks through the random-weight fixture, which takes 760 decode steps and crosses the encoder window 47 times. The transcript and token count must equal offline `generate`, and the session may never hold more than 8 conv rows and 24 adapter rows. The whole minute is 3 050 and 762.
- `stepAfterTheStreamFinishedBuffersNoSamples` ends a stream on `maxTokens`, keeps streaming, and expects no buffered samples.

## Measurements

All three PRs together against `8ed8188`. Voxtral Mini 4B Realtime on an M-series MacBook Pro, fed speech-band noise in 100 ms steps. Step time is the mean over the last second of audio, and memory is MLX active memory.

| Into the stream | Before: step | Before: memory | After: step | After: memory |
|---|---|---|---|---|
| 1 min | 62 ms | 2 744 MB | 61 ms | 2 757 MB |
| 10 min | 105 ms | 4 266 MB | 65 ms | 4 061 MB |
| 30 min | 115 ms | 4 870 MB | 81 ms | 4 217 MB |
| 60 min | 147 ms | 5 573 MB | 66 ms | 4 217 MB |
| 180 min | | | 66 ms | 4 217 MB |

The decoder window fills at 11 minutes. From there to three hours, memory stays at 4 217 MB and the step between 65 and 81 ms. Before, the step stayed above the 100 ms real-time budget from 15 minutes on, so live text fell further behind the speaker every minute. I did not bench the PRs separately. The transcript of a fixed speech passage is byte-identical before and after.

Not addressed: decoder RoPE angles are absolute position times frequency in float32, so their precision drops over multi-hour streams. The spacing reaches 1/64 radian after about 3 hours. I have not measured whether that changes any output.

CI: the full suite currently hangs the macOS runner on unmodified `main` (`3e97855`), so expect a red check here that is unrelated to this change. The VoxtralRealtime suites pass with the three PRs together: [30 tests in 6 suites](https://github.com/T0mSIlver/mlx-audio-swift/actions/runs/35285052551).


lucasnewman (approved):
Looks good, thanks!

PR #264, opened 2026-09-17
VoxtralRealtime decoder: append KV rows in place instead of rebuilding the window

T0mSIlver:
`VoxtralRealtimeDecoderKVCache` was a value that held exactly the sliding window. Every decoded token concatenated its keys and values onto the whole window and sliced off the oldest row, in every decoder layer, and the old and new copies stayed alive together until the step finished.

For Voxtral Mini 4B Realtime the window is 8 192 rows across 26 layers, about 870 MB at float16. Once it is full, every token copies all of it. While it filled, over the first 11 minutes of a stream, active memory rose 1.7 GB.

Independent of #263 and #265, which cover the rest of the long-stream work. #263 has the measurements for all three: at 10 minutes, when this cache is nearly full, a step takes 65 ms against 105 ms before.

## Changes

- `VoxtralRealtimeDecoderKVCache` is a class whose storage grows a 256-row block at a time. `append` writes the new rows with a slice update and returns what attention read before: the last `slidingWindow` rows and the absolute position of the first one.
- MLX runs the slice update in place when nothing else references the storage buffer. A second live reference would make every append copy the whole storage, which costs more than the old scheme. None of the callers (the stream session, `generate`, `generateStream`) holds one, and the class comment states the requirement.
- Rows that fall out of the window stay in place until a block of them has piled up or the storage is full. Then the window is copied down to row zero in one move. After the first fill the storage stops growing, at the window plus at most a block.
- The compaction copy can overlap itself. That is safe because mlx-swift's slice setter builds a lazy `slice_update` into a fresh output from the array's current handle, and the right-hand side is a slice of that same handle, so the source cannot be overwritten mid-copy.
- The index arithmetic is a separate value, `VoxtralRealtimeDecoderKVCacheAppendPlan`, so tests can cover it without Metal.

## Tests

- `VoxtralRealtimeDecoderKVCachePlanTests` checks growth at exactly one block, compaction only after a block of rows has piled up, compaction before growth when rows can be reclaimed, and a position offset equal to the old eager trim over 2 000 appends.
- `VoxtralRealtimeDecoderKVCacheTests` runs real arrays against the old scheme. A 5-row prefill and 1 500 single-row appends with a 300-row window go through one growth and several overlapping compactions. After every append the returned keys, values and position offset must equal concatenate-then-trim exactly.

The random-weight fixture used by the streaming tests has no decoder layers, so those tests never reach this cache. With the real model, the transcript of a fixed speech passage is byte-identical before and after.

CI: the full suite currently hangs the macOS runner on unmodified `main` (`3e97855`), so expect a red check here that is unrelated to this change. The VoxtralRealtime suites pass with the three PRs together: [30 tests in 6 suites](https://github.com/T0mSIlver/mlx-audio-swift/actions/runs/35285052551).


lucasnewman (approved):
Looks good!

PR #265, opened 2026-09-17
VoxtralRealtime streaming: decode the transcript one token at a time

T0mSIlver:
After every step the stream session decoded its whole token list back into text, then compared that text with the previous step's using `hasPrefix` and `dropFirst`. Both cost time proportional to the transcript. The session decodes 12.5 tokens per second of audio, padding included, so after an hour every step walked 45 000 tokens and about 50 KB of text.

#263 has the measurements for the three long-stream PRs together.

## Changes

- New `VoxtralRealtimeTranscriptText` appends each kept token's bytes as the token is decoded. It holds the finished text plus the first bytes of a character whose remaining bytes have not arrived, at most three.
- Bytes only move into the finished text where a one-pass UTF-8 decoder has finished a character. A lead byte always starts a new character, and a sequence with all its continuation bytes reads the same whatever follows. So `text` stays equal to `String(decoding: allBytes, as: UTF8.self)`, replacement characters included.
- `Delta.text` keeps its meaning: the new suffix when the old transcript is a prefix of the new one, and the whole new transcript when it is not. That happens when a trailing U+FFFD becomes a real character, or when new text combines with the last character, for example a combining accent. Only the last finished character and what follows it can compare differently, so only that tail is compared.
- The session appends bytes for every token `generated` keeps, including the one that crosses `maxTokens`, and skips a trailing EOS, as the whole-list decode did.
- `streamingTokenBytes(_:)` replaces `decodeStreaming(_:)`, and the tokenizer's per-token lookup `tokenBytes(for:)` becomes internal.
- While a character is incomplete, reading `text` builds a new string as long as the transcript. The old stored property never copied. The doc comment points hosts that only need each step's new text to `Delta.text`.

## Tests

- `matchesOnePassDecodingAndWholeTranscriptDeltas` runs 64 seeded sequences of 200 steps. Each step appends one to three tokens from a pool that splits two-, three- and four-byte characters at every position, adds combining accents, regional indicators and a zero-width joiner, and includes bytes that are never valid. After every step `text` must equal a one-pass decode and the delta must equal the old whole-transcript comparison, scalar for scalar.
- `holdsBackOnlyAnUnfinishedTrailingSequence` checks the cut rule on hand-picked byte sequences.
- `textEndsInAReplacementCharacterWhileACharacterIsIncomplete` and `aDeltaWithNoNewBytesIsEmpty` cover the two edge cases callers see.
- `splitMultibyteTokensMatchOfflineTranscript` streams 20 seconds through the random-weight fixture with a vocabulary whose tokens split "é", "€" and "😀" into separate bytes. The transcript must equal offline `generate`, which still decodes all tokens in one pass.

With the real model, the transcript of a fixed speech passage is byte-identical before and after.


lucasnewman:
@T0mSIlver Thanks! The change looks good but the test conflicts need resolution.

T0mSIlver:
@lucasnewman Rebased onto main as a single commit now that #263 and #264 are in; conflicts resolved and CI is green.

lucasnewman (approved):
Thanks!

PR #266, opened 2026-09-18
Qwen3ASR: pad mel chunks to the fixed chunk size, not the batch max

mattmireles:
## Summary

`Qwen3ASRAudioEncoder` pads each mel chunk to `maxChunkLen` — the largest chunk length *present in the batch* — instead of to the fixed `chunkSize` (`n_window * 2`, i.e. 100 frames). For any clip containing at least one full chunk this is already 100 and behaviour is unchanged. For a clip **shorter than 1.00 s** there is only one short chunk, so the conv stem convolves it at its own length and diverges from the reference implementation.

Two sites, both changed here:

- `Qwen3ASRAudioEncoder.callAsFunction` — the batched path
- `Qwen3ASRAudioEncoder.encodeSingleWindow` — the streaming path

## Why the reference always pads to a full chunk

In 🤗 `transformers` (`models/qwen3_asr`):

- `Qwen3ASRFeatureExtractor.__call__` right-pads the mel time axis with zeros to a multiple of `n_window * 2`, and the returned mask keeps the **unpadded** frame count (`feature_extraction_qwen3_asr.py`).
- `Qwen3ASREncoder.forward` raises unless `padded_feature_length % (n_window * 2) == 0`, then splits with a plain `view`/`reshape` into fixed-width tiles — there is no partial-chunk path (`modeling_qwen3_asr.py`).
- The true length re-enters only *after* the conv stem, as a gather keeping the first `_post_cnn_length(F)` of the 13 positions each chunk produces.

Verified with a forward hook: for a 50-frame (0.50 s) clip the reference hands `conv2d1` a `(1, 1, 128, 100)` tensor. Since the feature extractor pads to a multiple of 100 at training time, full-width chunks are training-time semantics.

## Impact

Measured on a fine-tuned Qwen3-ASR checkpoint, fp32, comparing conv-stem output of the current behaviour against the reference:

| Mel frames | Clip | Kept rows | Cosine | Max abs err | Ref abs max |
|---:|---:|---:|---:|---:|---:|
| 10 | 0.10 s | 2 | 0.98262 | 1.293 | 7.84 |
| 25 | 0.25 s | 4 | 0.99501 | 0.931 | 9.66 |
| 50 | 0.50 s | 7 | 0.99516 | 1.280 | 10.02 |
| 75 | 0.75 s | 10 | 0.99685 | 1.499 | 9.84 |
| 99 | 0.99 s | 13 | **1.00000** | **0.0** | 9.68 |

The 0.99 s row is the control: once no kept position's receptive field reaches the padding boundary the two agree exactly, which is what confines this to sub-second input.

End-to-end greedy decoding over 41 sub-second clips produced byte-identical transcripts 39/41 (95.1%); the 2 disagreements were genuine word changes, not punctuation. A control set of 32 clips in the 1.2–3.0 s range was 32/32 identical, confirming the change is inert above one chunk.

## Why this is safe

- **Output row count is unaffected.** Kept rows come from `getFeatExtractOutputLengths(clen)` on each chunk's *true* length, and each chunk is sliced with `x[i, 0..<validLen]`. `maxChunkLen` never entered that computation, so the audio-token count cannot change.
- **Positional embeddings are unaffected.** `Qwen3ASRSinusoidalPE` returns `_positionalEmbedding[0..<seqLen]`, a prefix slice of a precomputed table, so widening the conv leaves the kept rows' embeddings bit-identical.
- **No-op above one chunk.** Any input with at least one full chunk already had `maxChunkLen == 100`.

The change is a local-variable rename plus a comment; `chunkSize` is already in scope at both sites.


lucasnewman:
@mattmireles Thanks for the contribution! Your fix looks correct but the prompt placeholders need similar trimming. I put up a change in https://github.com/Blaizzy/mlx-audio-swift/pull/269 that should cover it.

